PR #17, opened 2018-06-28
add extern c to axl.h

adammoody:


adammoody:
So that C++ codes can include AXL as simply:

#include "axl.h"

rather than:

extern "C" {
#include "axl.h"
}

PR #19, opened 2019-02-20
Fixes for bbapi create

kshoga1:
Fixing the order of creating the definition and handle for bbapi. Also adding in error checking.


gonsie:
looks great! Thanks!

PR #20, opened 2019-03-08
Get IBM BB test working

tonyhutter:
- Get IBM burst buffer test working.
- Add command line arguments to test_axl.
- More extensive verification of destination file after transfer.
- Use the number at the end of the hostname for our "rank" that gets passed to BB_InitLibrary().
- Use seconds + nanoseconds to create our unique BBTAG.
- Misc cleanups
- Use 4-space indents to match the rest of the files.

### DIFF
--- a/src/axl_async_bbapi.c
+++ b/src/axl_async_bbapi.c
@@ -1,9 +1,13 @@
+#include <assert.h>
+#include <errno.h>
+#include <time.h>
+#include <sys/types.h>
+#include <sys/syscall.h>
 #include "axl_internal.h"
 #include "axl_async_bbapi.h"
 
 #ifdef HAVE_BBAPI
 #include <bbapi.h>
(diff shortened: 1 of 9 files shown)


gonsie:
Looks great!

Some of the changes in the tests don't quite make sense to me, but if it works, it works.

adammoody:
Thanks @tonyhutter !  We can also ask IBM for a review during a call sometime.  They should be able to help fill in details about the API where things aren't clear to us.

tonyhutter:
@gonsie I added the `{}` brackets in my latest push

tonyhutter:
In my latest push, I changed the unique BBTAG code to be:

`(unix timestamp << 32 | thread ID)`

This should be fine unless the user wraps the TID in the same second.  In order to do that they would have to spawn more than `/proc/sys/kernel/pid_max` AXL-using threads in the same second (currently 180k+ on my development system).

Issue #21, opened 2019-03-19
warning: implicit declaration of function ‘fprintf’


tonyhutter:
Today both @CamStan and I hit a build error caused by https://github.com/ECP-VeloC/AXL/pull/20:
```
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:71:9: warning: implicit declaration of function ‘fprintf’ [-Wimplicit-function-declaration]
         fprintf(stderr, "Hostname too long\n");
         ^~~~~~~
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:71:9: warning: incompatible implicit declaration of built-in function ‘fprintf’
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:71:9: note: include ‘<stdio.h>’ or provide a declaration of ‘fprintf’
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:71:17: error: ‘stderr’ undeclared (first use in this function); did you mean ‘opterr’?
         fprintf(stderr, "Hostname too long\n");
                 ^~~~~~
                 Opterr
⋮
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c: At top level:
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:131:8: error: unknown type name ‘BBTAG’
static BBTAG axl_get_unique_tag(void)
        ^~~~~
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c: In function ‘axl_get_unique_tag’:
prefix/VELOC/deps/AXL/src/axl_async_bbapi.c:135:5: error: unknown type name ‘uint64_t’; did you mean ‘u_int64_t’?
     uint64_t timestamp;
     ^~~~~~~~
     u_int64_t
```
Strangely, I don't hit it when building with the IBM BB API:
```
$ make
Scanning dependencies of target axl_o
[  8%] Building C object src/CMakeFiles/axl_o.dir/axl_async_bbapi.c.o
[ 66%] Built target axl_o
Linking C shared library libaxl.so
[ 66%] Built target axl
Linking C static library libaxl.a
[ 66%] Built target axl-static
Linking C executable test_axl
[100%] Built target test_axl
[hutter2@butte5:build]$ make clean
[hutter2@butte5:build]$ make
[  8%] Building C object src/CMakeFiles/axl_o.dir/axl.c.o
[ 16%] Building C object src/CMakeFiles/axl_o.dir/axl_sync.c.o
[...]

PR #22, opened 2019-03-20
Fix compile errors post-PR-20

tonyhutter:
#20 introduced build errors when not compiling using the IBM BB API. This patch fixes the issue.

Fixes: #21


adammoody:
Thanks @tonyhutter !

PR #23, opened 2019-03-20
Make BB API optional and autodetect headers in /opt/ibm/bb

tonyhutter:
- Make the BB API optional.  It will get built if the headers and libs are available.
- Autodetect BB API headers and libraries in `/opt/ibm/bb` if `BBAPI_LIBRARIES` and `BBAPI_INCLUDE_DIRS` are not specified.


adammoody:
Thanks, @tonyhutter .  Looks good to me.  This ready to go?

tonyhutter:
@adammoody it correctly detects the BB API without manually specifying BBAPI_LIBRARIES and BBAPI_INCLUDE_DIRS.  I haven't tried running `text_axl` with it, since our test machine is down, but it's a build-only change so it should be fine.

Issue #25, opened 2019-03-20
Failed build on system w/out BB API after PR-23


CamStan:
After #23, when building on a system w/out IBM BB API I get an error:
```
-- Found ZLIB: /usr/lib64/libz.so (found version "1.2.7")
CMake Error at /usr/tce/packages/cmake/cmake-3.9.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find BBAPI (missing: BBAPI_LIBRARIES BBAPI_INCLUDE_DIRS)
Call Stack (most recent call first):
  /usr/tce/packages/cmake/cmake-3.9.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:377 (_FPHSA_FAILURE_MESSAGE)
  cmake/FindBBAPI.cmake:22 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:64 (FIND_PACKAGE)

-- Configuring incomplete, errors occurred!
```
I'm still new to cmake, but do we need something like `IF(BBAPI_FOUND)` and `ENDIF(BBAPI_FOUND)` around these lines?

https://github.com/ECP-VeloC/AXL/blob/f3f79003335b428c7ae603bd7780d0bc9cf99ccd/CMakeLists.txt#L65-L68

PR #26, opened 2019-03-20
Fix BBAPI autodetect on systems w/out BBAPI

CamStan:
#23 introduced an error when attempting to autodetect BBAPI on systems without it. This patch fixes the issue.

Tested this on a system with and without BBAPI.

Fixes: #25


CamStan:
I'm stealing your tricks @tonyhutter...

 I didn't know about the "Fixes: #" feature until I saw #22.

adammoody:
Nice catch, @CamStan

tonyhutter:
Thanks for the fix!

PR #4, opened 2018-02-22
Initial attempt at travis

becker33:
Travis will build KVTree, build AXL against KVTree, and run AXL tests

Includes commented-out hook for adding codecov once we go public.

PR #29, opened 2019-03-25
Setup Travis CI

CamStan:
This sets up Travis CI testing and builds on multiple compilers that are on our machines. Can always scale back if it's overkill.

This will allow PR #4 to be closed since @becker33 has abandoned us.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,89 @@
+langauge: c
+dist: trusty
+
+compiler:
+  - gcc
+  - clang
+
+matrix:
+  include:
+  - name: "GCC 4.9"
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - g++-4.9
+    env :
+      - MATRIX_EVAL="CC=gcc-4.9 && CXX=g++-4.9"
+
+  - name: "GCC 6"
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - g++-6
+    env :
+      - MATRIX_EVAL="CC=gcc-6 && CXX=g++-6"
+
+  - name: "GCC 7"
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - g++-7
+    env :
+      - MATRIX_EVAL="CC=gcc-7 && CXX=g++-7"
+
+  - name: "GCC 8"
+    addons:
+      apt:
+        sources:
+          - ubuntu-toolchain-r-test
+        packages:
+          - g++-8
+    env :
+      - MATRIX_EVAL="CC=gcc-8 && CXX=g++-8"
+
+  - name: "Clang 4.0"
+    addons:
+      apt:
+        sources:
+          - llvm-toolchain-trusty-4.0
+        packages:
+          - clang-4.0
+    env :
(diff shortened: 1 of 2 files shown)


tonyhutter:
I love the multiple compiler targets!  Awesome!

1. Do you need `-DMPI=ON`?:
https://travis-ci.com/ECP-VeloC/AXL/jobs/187585322#L795

2. Can you have the builds fail on warnings?:
https://travis-ci.com/ECP-VeloC/AXL/jobs/187585322#L875

.travis.yml
```diff
@@ -0,0 +1,101 @@
...
+  - mkdir $HOME/KVTree/kvtree-build
+  - mkdir $HOME/KVTree/kvtree-install
+  - cd $HOME/KVTree/kvtree-build
+  - cmake .. -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_INSTALL_PREFIX=$HOME/KVTree/kvtree-install -DMPI=ON
+  - make
+  - make install
+
+script:
+  - cd $TRAVIS_BUILD_DIR
+  - cmake -DCMAKE_C_COMPILER=$CC -DCMAKE_CXX_COMPILER=$CXX -DCMAKE_C_FLAGS="-coverage" -DMPI=ON -DAXL_ASYNC_DAEMON=OFF -DWITH_KVTREE_PREFIX=$HOME/KVTree/kvtree-install
```
CamStan: Hmm, I do have `-DMPI=ON` but
```
CMake Warning:
  Manually-specified variables were not used by the project:

    MPI
```
still shows up. This shows up when I build locally as well, but only for AXL.

tonyhutter: Yea, I think we can get rid of it

CamStan:
I've never had CI set up to fail on warnings, but it might be possible if we set the `-Werror` C_FLAG. I'd have to test that if that's the route we want to go.

tonyhutter:
My vote would be for `-Wall -Werror`

adammoody:
I don't think AXL has an MPI option, since it doesn't use MPI.  Also, we probably don't need to enable MPI in the kvtree build for AXL either if you want to leave that out.

CamStan:
Pulled out MPI since AXL doesn't use it and KVTree's Travis already tests with it.

I added `-Wall -Werror` and started cleaning up the errors to get it to pass, but ran into an issue here https://travis-ci.org/ECP-VeloC/AXL/jobs/511226558#L723 where Travis complains the function is unused.

Essentially, `axl_async_daemon()` only gets called by other functions inside of `#ifdef HAVE_DAEMON` blocks which in this case is turned off.

https://github.com/ECP-VeloC/AXL/blob/d427c238d0f24dd874a7deb852f219dc79695376/src/axl_async_daemon.c#L335-L337

Any thoughts?

CamStan:
Or would it be better to leave `-Wall -Werror` out for now so that we can get Travis at least building all PR's? Then we can fix those warnings in other PR's and then put those flags back in?

tonyhutter:
> Or would it be better to leave -Wall -Werror out for now so that we can get Travis at least building all PR's? 

I'm fine with that

CamStan:
Issue #31 addresses the warnings for turning on `-Werror`.

Otherwise this should be good to go now.

adammoody:
Nice!  Thanks @CamStan and @tonyhutter .

PR #30, opened 2019-03-25
Add AXL_XFER_BEST to autodetect the best transfer API

tonyhutter:
Add a new `AXL_XFER_BEST` type for `AXL_Create()`.  This allows the caller to leave it up to AXL to choose the best API for the particular node.  For example, if you're running on an IBM node, use the BB API. If you're running on a Cray, use DataWarp.  Otherwise use sync.  That way the caller doesn't have to have hardware-specific transfers in their checkpointing code.

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -90,6 +90,33 @@ static int axl_get_info(int id, kvtree** list, axl_xfer_t* type, axl_xfer_state_
 }
 
 /*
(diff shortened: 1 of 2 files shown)


tonyhutter:
Note: we could modify this to detect the best API at runtime.  For example, if BBPATH is set, then use the BBAPI.  This could be useful for cases where the BB server is down.

PR #32, opened 2019-03-26
Bump version to 0.2.0

tonyhutter:
Tag version 0.2.0 to mark API change for addition of `AXL_XFER_BEST`. That way, we can point VeloC at the right version of AXL when we update it to use `AXL_XFER_BEST`.

Also, remove unused `AXL_MAJOR`, `AXL_MINOR`, and `AXL_PATCH`.


tonyhutter:
I just pushed out the tag

PR #35, opened 2019-05-15
Add "default" and "native" transfer types

tonyhutter:
This adds in `AXL_XFER_DEFAULT` and `AXL_XFER_NATIVE`, and removes `AXL_XFER_BEST`:

AXL_XFER_DEFAULT:
Let AXL choose the fastest transfer type that is compatible with all VeloC transfers.  This may or may not be the node's native transfer library.

AXL_XFER_NATIVE:
Use the node's native transfer library (like IBM Burst Buffer or Cray DataWarp) for transfers.  These native libraries may or may not support all VeloC
transfers.

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -57,7 +57,6 @@ If a transfer fails, partially transferred files are not removed
 from the destination.
 
 ## Transfer types
-Currently, there are four transfer types:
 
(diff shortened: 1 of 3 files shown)


tonyhutter:
@adammoody braces fixed in my latest push

tonyhutter:
I updated `doc/README.md` with the new xfer types in my latest push.

PR #36, opened 2019-05-21
Bump version to 0.3.0

tonyhutter:
Tag version 0.3.0 to mark API change for addition of AXL_XFER_DEFAULT and AXL_XFER_NAVIVE, and the removal of AXL_XFER_BEST.

After pulling in this commit, please create an `v0.3.0` tag.

### DIFF
--- a/src/axl.h
+++ b/src/axl.h
@@ -8,7 +8,7 @@ extern "C" {
 
 #define AXL_SUCCESS (0)
 
-#define AXL_VERSION "0.2.0"
+#define AXL_VERSION "0.3.0"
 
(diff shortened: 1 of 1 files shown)


tonyhutter:
I rebased this on master.  You should be able to pull it in now without issue.

tonyhutter:
Can you rename the tag from `0.3.0` to `v0.3.0`?  That `v` seems to be the convention for the ECP-VeloC projects.

gonsie:
Added the v... but I think I have to go in and delete the old tag by hand. Hopefully tomorrow.

PR #42, opened 2019-05-23
Add pthreads xfer, axl_cp, and update AXL_Add for directories

tonyhutter:
- Add a new `AXL_XFER_PTHREAD` transfer.  This uses multiple threads to transfer the files.

- Add an `axl_cp` command.  `axl_cp` works like the standard `cp` command, but allows you to specify the AXL transfer type (sync, pthreads, bbapi, etc...). It will also serve as good example code for AXL users.  Along with this, `test_axl.c` has been replaced with a `test_axl` bash script.  This will make it a lot easier to write test cases.  The new `test_axl` script will test both sync and pthread transfers via calls to `axl_cp`.

- Allow passing directories to `AXL_Add()`.  If a directory is passed, all files and subdirs within the dir are recursively copied.  Basically, `AXL_Add()` will follow the same semantics as the `cp` command.  This is conceptually easier for the user to understand, and saves them from having to traverse their directories manually.  Note that the recursive directory copy will currently not copy empty subdirectories.


tonyhutter:
Travis caught this:
`../src/libaxl.so: undefined reference to 'pthread_join'`
Looks like I forgot to update CMakeLists to include `-lpthread`.  Let me fix that...

test/axl_cp.c
```diff
@@ -0,0 +1,195 @@
+#include <stdlib.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <string.h>
+#include <limits.h>
+#include <errno.h>
+#include <sys/stat.h>
+#include <signal.h>
+#include "axl.h"
```
CamStan: Looks like the errors are happening when building `axl_cp.c`. Do you need to include `pthread.h` or `axl_pthread.h` here?

test/CMakeLists.txt
```diff
@@ -3,21 +3,18 @@
 ###############
 
 LIST(APPEND axl_test_srcs
-    test_axl.c
-    test_axl_sync.c
-    test_axl_async_daemon.c
-    test_axl_async_ibmbb.c
+    axl_cp
```
CamStan: Is the `.c` needed on the end of this?

tonyhutter: It builds for me without it, which seems weird.  I'll add it in

tonyhutter:
Travis is hitting a `make check` failure when running `axl_cp`.  I'll need to update that too

tonyhutter:
I needed a `make check` test case for Travis to run, so I wrote a `test_axl` script which invokes `axl_cp`.  I've included the `test_axl` script in my latest push.  I've also updated my PR comments to reference it.

`make check` works in my Fedora 29 VM, but it's failing on Travis, so I need to dig into that.

tonyhutter:
Looks like Travis is erroring out here:
```
1: Test command: /home/travis/build/ECP-VeloC/AXL/test/test_axl
1: Test timeout computed to be: 9.99988e+06
1: Testing sync transfer...	failed copy
1: AXL 0.3.0 ERROR: travis-job-316ad4c6-c516-4ece-a8ba-645769a85db5: axl_async_init_daemon: child process failed to launch: axld /dev/shm/axld: No such file or directory @ axl_async_init_daemon /home/travis/build/ECP-VeloC/AXL/src/axl_async_daemon.c:578
1: AXL_Init() failed (error -1)
1: AXL 0.3.0 ERROR: travis-job-316ad4c6-c516-4ece-a8ba-645769a85db5: axl_async_init_daemon: child process failed to start: axld /dev/shm/axld @ axl_async_init_daemon /home/travis/build/ECP-VeloC/AXL/src/axl_async_daemon.c:601
1: AXL_Init() failed (error -1)
1/1 Test #1: transfer_tests ...................***Failed    5.67 sec
```
I can reproduce this on my VM when I build with `-DAXL_ASYNC_DAEMON=ON` like Travis does.  It appears to be an existing bug that's unrelated to my change, as I get the same error when running `test_axl -s` in the `master` branch:
```
$ ./test/test_axl -s
Using existing source file /tmp/testfile
AXL 0.3.0 ERROR: fedora29: axl_async_init_daemon: child process failed to launch: axld /dev/shm/axld: No such file or directory @ axl_async_init_daemon /home/hutter/AXL-master/src/axl_async_daemon.c:578
AXL 0.3.0: fedora29: axl_sync_start: Read and copied test_file to /home/hutter/AXL-master/build/test_file_moved with success code 0 @ axl_sync_start /home/hutter/AXL-master/src/axl_sync.c:46
```
The only reason that Travis is passing on `master` is that `make check` is incorrectly calling `test_axl` with no arguments, and so no tests get run, and thus this `axl_async_init_daemon()` code is never invoked.

Since this is broken, I'm going to remove `-DAXL_ASYNC_DAEMON=ON` from `.travis.yml`.

CamStan:
It's not building for me when I pull your branch. Tagging @gonsie to review the cmake.

tonyhutter:
@CamStan can you pull again and try building?  If it's still not working, can you post the build failure?

CamStan:
It built for me this time. Might have been something on my end that cause my failure last week.

test/CMakeLists.txt
```diff
@@ -3,22 +3,24 @@
...
-    test_axl_async_daemon.c
-    test_axl_async_ibmbb.c
+    axl_cp.c
 )
 
-ADD_EXECUTABLE(test_axl ${axl_test_srcs})
-TARGET_LINK_LIBRARIES(test_axl axl)
+ADD_EXECUTABLE(axl_cp ${axl_test_srcs})
+
+find_package(Threads REQUIRED)
```
gonsie: The tests shouldn't need to find pthreads unless the test code itself is using the pthread library.

tonyhutter: Good catch.  I'll fix it in my next push

gonsie: Same with the `TARGET_LINK_LIBRARIES`, the test should get the link to pthread through linking to AXL. If that isn't working we have to change the cmake in src.

CMakeLists.txt
```diff
@@ -43,6 +43,12 @@ IF(ZLIB_FOUND)
 	LIST(APPEND AXL_LINK_LINE "-lz")
 ENDIF(ZLIB_FOUND)
 
+# PTHREADS
+FIND_PACKAGE(Threads REQUIRED)
+IF(THREADS_HAVE_PTHREAD_ARG)
+	LIST(APPEND AXL_LINK_LINE "-lpthread")
```
gonsie: Use `AXL_EXTERNAL_LIBS` instead. The cmake’s findThreads module is documented here: https://cmake.org/cmake/help/v3.14/module/FindThreads.html which lists what variables it is setting. I think we want to use the `_INIT` variable??

tonyhutter:
@gonsie thanks `AXL_EXTERNAL_LIBS` did the trick.  It's building now.

src/axl_pthread.c
```diff
@@ -0,0 +1,369 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <sys/sysinfo.h>    /* get_nprocs() */
```
gonsie: can we use something other than sys/sysinfo? Mac's don't have that.

tonyhutter: Maybe?:
```
#include <unistd.h>
long number_of_processors = sysconf(_SC_NPROCESSORS_ONLN);
```
One commenter says it works on Linux and OSX:
https://stackoverflow.com/questions/4586405/how-to-get-the-number-of-cpus-in-linux-using-c/4586471#4586471

gonsie:
@tonyhutter how do I submit changes back to your branch? I'm not familiar with the all-in-one-commit model.

tonyhutter:
Hmm.. I don't know if you can push to my branch.  You can post you patch directly in a comment and I can apply it.  That way all the reviewers can see your proposed change.  I've done that in the past for some ZFS code reviews.

tonyhutter:
Alternatively, you can create a branch with your patch in it and point me to it.

gonsie:
```
From 186e5e0dfed377d5d89733541c0fc13b1613d531 Mon Sep 17 00:00:00 2001
From: "Elsa Gonsiorowski (Uranus)" <gonsiorowski1@llnl.gov>
Date: Fri, 31 May 2019 10:11:43 -0700
Subject: [PATCH 1/2] get_nprocs on apple

---
 src/axl_pthread.c | 19 ++++++++++++++++++-
 1 file changed, 18 insertions(+), 1 deletion(-)

diff --git a/src/axl_pthread.c b/src/axl_pthread.c
index 3a2950b..d128f84 100644
--- a/src/axl_pthread.c
+++ b/src/axl_pthread.c
@@ -1,11 +1,18 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <pthread.h>
-#include <sys/sysinfo.h>    /* get_nprocs() */
 #include "axl_internal.h"
 #include "kvtree_util.h"
 #include "axl_sync.h"
 
+/* get_nprocs() */
+#if defined(__APPLE__)
+#include <sys/sysctl.h>
+#else
+#include <sys/sysinfo.h>
+#endif
+
+
 #define MIN(a,b) (a < b ? a : b)
 
 /*
@@ -200,7 +207,17 @@ int axl_pthread_start (int id)
     file_count = kvtree_size(files);
 
     /* Get number of hardware threads on our CPU (includes HT) */
+#if defined(__APPLE__)
+    int count;
+    size_t size = sizeof(count);
+    if (sysctlbyname("hw.ncpu",&count,&size,NULL,0)) {
+        cpu_threads = 1;
+    } else {
+        cpu_threads = count;
+    }
+#else
     cpu_threads = get_nprocs();
+#endif
 
     threads = MIN(cpu_threads, MIN(file_count, MAX_THREADS));
 
-- 
2.17.2 (Apple Git-113)

```

gonsie:
Could you also change the test_axl script name to test_axl.sh?

tonyhutter:
@gonsie thanks for the patch!  I'll apply it and retest.  I'll also do the `test_axl.sh` rename.

gonsie:
The test is failing on my machine... the temp files that are created cannot be opened by the AXL library (even though they seem to exist). But, I'll stop fiddling with this PR and approve it. We can work on testing later.

tonyhutter:
@gonsie I included your changes in my latest push.

gonsie:
looks like the patch I submitted didn't apply cleanly and the sys/sysinfo.h include is still in there. FYI, I just pushed a fix to master.

Issue #44, opened 2019-05-31
BB_AddFiles() checks if destination path exists (need workaround)


tonyhutter:
If you `AXL_Add()` a file with a destination path that doesn't exist, it's totally fine, since AXL will automatically create the path for you in `AXL_Dispatch()`.  So if you do:

`AXL_Add(id, "/tmp/myfile", "/tmp/dir1/dir2/dir3/myfile")`

`AXL_Dispatch()` will mkdir `/tmp/dir1/dir2/dir3/` before copying the file.  

This is an issue with the BB API, since it checks to see if the destination path exists in `BB_AddFiles()` (a.k.a `AXL_Add()`).  A possible solution to this to do both the `BB_AddFiles()` of the entire file list, and the `BB_StartTransfer()`  in `axl_async_start_bbapi()` (a.k.a `AXL_Dispatch()`).  That way we guarantee the directories are there before we start copying the files.

PR #45, opened 2019-05-31
 Do file list adds during AXL_Dispatch for BB API

tonyhutter:
The BB API checks to see if the destination path exists at BB_AddFiles() time (analogous to AXL_Add()).  This is an issue since the destination paths get mkdir'd in AXL_Dispatch() and thus aren't available yet.  This patch holds off on doing our BB_AddFiles() calls until AXL_Dispatch().

Fixes: #44

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -363,7 +363,14 @@ __AXL_Add (int id, const char* src, const char* dest)
     case AXL_XFER_ASYNC_DW:
         break;
     case AXL_XFER_ASYNC_BBAPI:
-        rc = axl_async_add_bbapi(id, src, dest);
+        /*
+         * Special case:
+         * The BB API checks to see if the destination path exists at
+         * BB_AddFiles() time (analogous to AXL_Add()).  This is an issue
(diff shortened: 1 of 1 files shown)


tonyhutter:
Rebased on top master

src/axl.c
```diff
@@ -526,6 +533,16 @@ int AXL_Dispatch (int id)
...
         axl_mkdir(dest_dir, mode_dir);
+
+        /*
+         * Special case: The BB API checks if the destination path exists at
+         * its equivalent of AXL_Add() time.  That's why we do its "AXL_Add"
+         * here, after the full path to the file has been created.
+         */
+        if (xtype == AXL_XFER_ASYNC_BBAPI) {
+            char *src = kvtree_elem_key(elem);
+            axl_async_add_bbapi(id, src, dest);
```
gonsie: should we check the return code here? or does the axl_async_add_bbapi throw up an error if one is encountered?

tonyhutter: It will print an error via a failed `bb_check()`, but won't stop anything.  I just added a check in my latest push that will catch the error though,

PR #47, opened 2019-05-31
Remove the AXL daemon

tonyhutter:
With the pthreads implementation added, we no longer need the AXL daemon.


tonyhutter:
Rebased on master and removed commit dependencies

gonsie:
I think that `src/axld.c` should also be removed.

tonyhutter:
Whoa, how did I miss that!  Just removed it in my latest push

PR #54, opened 2019-06-06
Enable -Wall -Werror by default, fix compiler warnings

tonyhutter:
- Enable `-Wall` and `-Werror` by default.  Enable `-Werror` in Travis.

- Fix all compiler warnings:
```
axl_async_bbapi.c:65:12: warning: ‘axl_get_unique_node_id’ defined but not used
axl.c:605:25: warning: unused variable ‘bytes_written’
axl.c:605:12: warning: unused variable ‘bytes_total’
```


CamStan:
Apparently this:
```
/home/travis/build/ECP-VeloC/AXL/test/axl_cp.c:89:32: error: suggest braces
      around initialization of subobject [-Werror,-Wmissing-braces]
    struct sigaction action = {0};
                               ^
                               {}
```
was considered a bug in GCC 4.9 and they "fixed" it ([here](https://stackoverflow.com/questions/11551986/gcc-missing-braces-around-initializer)), although with some ongoing discussion around it.
Some also consider it a bug in Clang ([here](https://stackoverflow.com/questions/31555584/why-is-clang-warning-suggest-braces-around-initialization-of-subobject-wmis)), while others suggest that this:
https://github.com/ECP-VeloC/AXL/blob/7e60afd733dd07d9efb9e7cd5587e97c5510a611/test/axl_cp.c#L89
should be
```
struct sigaction action = {{0}};
                          ^   ^
```
in C.

tonyhutter:
@CamStan thanks for the info.  I just switched it to a `memset()` in my latest push to be super compatible with everything.

Issue #57, opened 2019-06-10
Expected behavior of AXL after being killed


tonyhutter:
Let's say you SIGKILL or crash an app in the middle of an AXL transfer, and then reload it with the same state file (via AXL_Init), what is the expected behaviour?  Should the failed transfer be resumed as part of the call to AXL_Init?  Should AXL_Init return an error and do nothing?

I started writing a test case for this but didn't know what to expect.

gonsie:
from today's meeting, if AXL Init notices that it is restarting (i.e., a previous state file exists) it should do nothing / have no smarts. It would be up to the orchestrating layer above AXL to call stop or resume.

tonyhutter:
So basically applications calling AXL should do this:
```
<app is running>
<crash>
<re-run app>
    AXL_Init(state_file)
    id = AXL_Create(xtype, name) //should give you back your old ID
    AXL_Cancel(id)
    AXL_Free(id)
```

gonsie:
For apps that didn't want to deal with ID's it could be:

```
AXL_Init(state_file);
AXL_Stop();
```

adammoody:
FYI, here is where SCRV stops flushes during SCR_Init:

SCR_Init() --> scr_flush_async_stop():
https://github.com/LLNL/scr/blob/82a3c43d8243c61fac331b64539535f6cdfa3961/src/scr.c#L1645

scr_flush_async_stop() --> Filo_Flush_stop():
https://github.com/LLNL/scr/blob/82a3c43d8243c61fac331b64539535f6cdfa3961/src/scr_flush_async.c#L37

Filo_Flush_stop() --> filo_axl_stop --> AXL_Stop:
https://github.com/ECP-VeloC/filo/blob/4c1bd39aae0208b17061a772ed05d80e07f38f58/src/filo.c#L953

tonyhutter:
Now that I'm digging into the code I see there's no way to get your old ID back.  So my example in https://github.com/ECP-VeloC/AXL/issues/57#issuecomment-501430547 doesn't work.  Here's what would happen:
```
<app is running with id = 1>
<crash>
<re-run app>
    AXL_Init(state_file)
    id = AXL_Create(xtype, name) //gives you a new id of 2
    AXL_Cancel(id)
    AXL_Free(id)
```
So in reality the application would have to remember it's `id` from the time it crashed, and do this:
```
    AXL_Init(state_file)
    AXL_Cancel(old_id)
    AXL_Free(old_id)
[...]

PR #58, opened 2019-06-10
Add a test case for AXL_Cancel'ing a transfer

tonyhutter:
- Make canceling pthread transfers actually work
- Add "number of files" (`-n`) and "seconds until cancel" (`-s`) options to `test_axl`.
- Add test case to cancel a pthreads and BB API transfer
- Fix some unused variable build warnings for BB API code


tonyhutter:
Note that this doesn't test resuming a failed transfer.  We need to resolve https://github.com/ECP-VeloC/AXL/issues/57 first.

src/axl_pthread.c
```diff
@@ -283,10 +288,11 @@ int axl_pthread_test (int id)
 {
     kvtree* file_list = kvtree_get_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
     int64_t ptr;
-    struct axl_pthread_data *pdata;
+    struct axl_pthread_data *pdata = NULL;
 
     kvtree_util_get_int64(file_list, AXL_KEY_PTHREAD_DATA, &ptr);
     pdata = (struct axl_pthread_data *) ptr;
+    assert(pdata);
```
gonsie: Can you explain these few lines to me (L291-L295)? When does this assert fail?

tonyhutter:
My latest push adds a test case for cancelling a BB API transfer as well.

src/axl_pthread.c
tonyhutter: It's there for sanity checking only.  It could be useful to catch if someone does an `AXL_Test()` before the pthread data is initialized.

gonsie: Hmmm. We should probably define the behavior of AXL if you call test before dispatching (or is it initializing?) a transfer. We probably don't want to die. I think there is a more graceful/helpful error response we could do here.

tonyhutter: Nevermind, looks like we're currently sanity checking AXL_Test to only check during a dispatched transfer.
```
int AXL_Test (int id)
{
...
    /* check that handle is in correct state to test */
    if (xstate != AXL_XFER_STATE_DISPATCHED) {
        AXL_ERR("Invalid state to test UID %d", id);
        return AXL_FAILURE;
    }
```

Issue #51, opened 2019-06-01
AXL_Stop with pthreads


adammoody:
There is a subtle issue that may arise with the intended use of AXL_Stop and pthreads, perhaps other AXL methods as well.

When given a path for AXL to write a state file, it will record its internal kvtree to this file at various points.

https://github.com/ECP-VeloC/AXL/blob/42ff1dd2678573abd7056125b7eee8151f79f407/src/axl.c#L563

If the file exists, it will also read that file back to initialize itself when starting up:

https://github.com/ECP-VeloC/AXL/blob/42ff1dd2678573abd7056125b7eee8151f79f407/src/axl.c#L172

The main point of this is so that AXL can record info about potentially ongoing transfers so that they can be terminated, even after the process that started the transfer may have died and been restarted.  Since the process that might be stopping or checking on the transfer is different from the process that started it, an issue can come up when storing memory addresses in this structure, since the memory address likely is invalid in the restarted process.

The particular sequence I'm thinking of right now:
1) application starts transfer, and kvtree is stored to file including memory address of data structures
2) application dies before transfer completes
3) application restarts, calls AXL_Init, AXL reads kvtree back from file
4) application calls AXL_Stop to kill any ongoing transfers
5) AXL_Stop calls AXL_Cancel / AXL_Wait / AXL_Free
6) AXL_Free tries to read and free invalid memory from old address in kvtree

There are likely many ways to fix it, but I don't currently have a suggestion.

tonyhutter:
Yea, a by-product of using a single `id` for everything that there's no easy way to pass around structs between functions.  I was hoping we could just pass `*axl_pthread_data` in the kvtree, but it sounds like that not going to work.  We could do something hacky and internally maintain a `id` to struct lookup list.  Or just manually NULL out the `*axl_pthread_data` when we read back the kvtree from a file.  I dunno the answer, but I think you bring up a good point.

tonyhutter:
We should also add a test case to cancel a transfer.

tonyhutter:
Fix out for review: https://github.com/ECP-VeloC/AXL/pull/59

tonyhutter:
Fixed in #59

PR #59, opened 2019-06-25
Don't put axl_pthread_data pointer in KVTree

tonyhutter:
This patch uses a linked list to lookup which `axl_pthread_data` is associated with each AXL ID.  Why put it in a linked list instead of just storing the pdata in the kvtree?  Because if we put it in a kvtree, and the app dies, the pdata pointer becomes stale, and would get erroneously freed as part of an `AXL_Stop()`.  Instead we use `axl_pthread_data_lookup()`, `axl_pthread_data_add()`, and `axl_pthread_data_remove()` to access the data.  This makes it so the pdata is ephemeral, only existing while the app is running.

### DIFF
--- a/src/axl_pthread.c
+++ b/src/axl_pthread.c
@@ -33,6 +33,15 @@ struct axl_work
 
 struct axl_pthread_data
 {
+    /* AXL ID associated with this data */
+    int id;
+
+    /* Number of threads */
+    unsigned int threads;
+
+    /* This struct is in a linked list */
+    struct axl_pthread_data *next;
+
     /* Lock to protect workqueue */
     pthread_mutex_t lock;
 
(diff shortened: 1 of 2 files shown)


tonyhutter:
Fixes https://github.com/ECP-VeloC/AXL/issues/51

tonyhutter:
The build failures are a KVTree bug (https://github.com/ECP-VeloC/KVTree/issues/27).  We'll need to get that fixed first before this goes in.

tonyhutter:
https://github.com/ECP-VeloC/KVTree/pull/28 should fix the build failures

src/axl_pthread.c
```diff
@@ -65,6 +94,77 @@ unsigned int axl_get_nprocs(void)
...
+/* Get the pthread data for a given AXL ID */
+struct axl_pthread_data *
+axl_pthread_data_lookup(int id)
+{
+    struct axl_pthread_data *pdata;
+    struct axl_pthread_data *ret = NULL;
+
+    pthread_mutex_lock(&axl_all_pthread_data.lock);
+
+    while ((pdata = axl_all_pthread_data.head)) {
```
adammoody: Do we need a pdata = pdata->next in this loop?

adammoody: Maybe?

pdata = axl_all_pthread_data.head;
while (pdata) {
  ...
  pdata = pdata->next;
}

tonyhutter: Yep, good catch!

tonyhutter:
@adammoody I just pushed a fix for `pdata = pdata->next`.  Can you take another look?

src/axl_pthread.c
```diff
@@ -65,6 +94,79 @@ unsigned int axl_get_nprocs(void)
...
+    pthread_mutex_unlock(&axl_all_pthread_data.lock);
+}
+
+void
+axl_pthread_data_remove(int id)
+{
+    struct axl_pthread_data *pdata, *prev = NULL;
+
+    pthread_mutex_lock(&axl_all_pthread_data.lock);
+    while ((pdata = axl_all_pthread_data.head)) {
```
adammoody: Need the same here?

tonyhutter: Sorry, yea, let me fix it there too

tonyhutter:
@adammoody I fixed the other `pdata = pdata->next`

Issue #61, opened 2019-12-11
BBAPI should fall back to SYNC if filesystem doesn't support extents


tonyhutter:
BBAPI requires a filesystem that supports extents.  If the file being transferred is not on an extents-capable filesystem (like NFS or ZFS) then it should fall back to a SYNC or PTHREAD transfer for that file.  This will become important when SCR eventually uses AXL to move files between its various hierarchies of directories.

PR #62, opened 2019-12-13
BBAPI fallback to sync if FS doesn't support extents

tonyhutter:
If the user requests a BBAPI transfer, and one or more of the source or destination paths in the file list aren't to a filesystem that BBAPI supports, then fallback to doing a pthreads transfer for all the files in the file list.  This gets around BBAPI giving you a "file not found" when it tries to transfer to/from a filesystem that doesn't support extents (like NFS).

Fixes:  #61

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -490,6 +490,8 @@ int AXL_Dispatch (int id)
     kvtree* file_list = NULL;
     axl_xfer_t xtype = AXL_XFER_NULL;
     axl_xfer_state_t xstate = AXL_XFER_STATE_NULL;
+    kvtree_elem *elem = NULL;
+    char *dest;
 
     if (axl_get_info(id, &file_list, &xtype, &xstate) != AXL_SUCCESS) {
         AXL_ERR("Could not find transfer info for UID %d", id);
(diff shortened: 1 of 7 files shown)


tonyhutter:
Reviewers: should this fall back to pthreads instead of sync?

src/axl_async_bbapi.c
```diff
@@ -362,3 +390,47 @@ int axl_async_cancel_bbapi (int id) {
...
+ * to the source or destination nor supporting extents (which BBAPI requires).
+ * If we're in fallback mode, we use a more compatible transfer method.
+ */
+int
+axl_bbapi_in_fallback(int id)
+{
+    kvtree* file_list = kvtree_get_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
+    int bbapi_fallback = 0;
+
+    if (kvtree_util_get_int(file_list, AXL_BBAPI_KEY_FALLBACK, &bbapi_fallback) != 0) {
```
adammoody: check against KVTREE_SUCCESS instead of 0

src/axl.c
```diff
@@ -504,38 +506,44 @@ int AXL_Dispatch (int id)
...
-         */
-        if (xtype == AXL_XFER_ASYNC_BBAPI) {
-            char *src = kvtree_elem_key(elem);
+    /*
+     * Special case: The BB API checks if the destination path exists at
+     * its equivalent of AXL_Add() time.  That's why we do its "AXL_Add"
+     * here, after the full path to the file has been created.
+     */
+    if (xtype == AXL_XFER_ASYNC_BBAPI) {
+
```
adammoody: kill this blank line

adammoody:
I'm fine with having this fallback to pthreads if you want.  Seems like that's closer to what the user might expect if they want this to be async.

tonyhutter:
Ok, I'm going to update it to use pthreads then.  I'll add in your other changes too.

src/axl_io.c
```diff
@@ -17,6 +17,12 @@
 #include <errno.h>
 #include <string.h>
 
+/* fiemap */
+#include <linux/fiemap.h>
+#include <sys/ioctl.h>
+#include <linux/fs.h>
+#include <libgen.h>
+
```
gonsie: Do these includes change which OSes we can support? If they only needed for the BBAPI check, maybe we can wrap them in a config-time option.

tonyhutter: They're standard headers that should be on any linux system.  I'll wrap it in `#if HAVE_BBAPI...` though, since BBAPI is the only thing that uses it.

tonyhutter:
@gonsie @adammoody my latest push addresses all your comments and falls back to pthreads instead of sync.

adammoody:
Thanks @tonyhutter .

PR #64, opened 2020-04-30
Add ENABLE_BBAPI_FALLBACK, check FS type on BBAPI transfers

tonyhutter:
Previously if you attempted to copy a file using BBAPI, and either the source and destination didn't support file extents, then it would automatically fallback to using pthreads.  This change removes
the fallback by default, and returns an error instead.  The fallback can be re-enabled by passing -DENABLE_BBAPI_FALLBACK to cmake.

This patch also checks the source and destination filesystem type against a whitelist instead of checking if they support extents. This was more effective, as there were some filesystems that BBAPI could not copy to/from even though they supported extents.


tonyhutter:
This patch will help with https://github.com/LLNL/scr/issues/163

adammoody:
Nice touch to make this an option.  Thanks, @tonyhutter .

Issue #41, opened 2019-05-22
Preserve file group, permission bits, and timestamps


adammoody:
If user applications are setting specific groups or permission bits on the source files, I'm guessing they'd like to have those properties preserved after a transfer.  We don't currently do this.

As for timestamps, I can imagine some apps might want to preserve those (use timestamp from the source files) and some apps might want to assign new timestamps when creating the destination files.

And then there are things like extended attributes, although that's probably much less of a concern (much smaller user base).

Not sure how any of the vendor APIs deal with this.

tonyhutter:
Maybe the default guarantee should be "AXL preserves whatever your transfer type preserves" (which we should document), but then give the option to explicitly say "preserve timestamps" or "preserve XATTRs" or whatever in `AXL_Config()`.  That would work for most cases.  It wouldn't work in the case where you wanted different files in your file list to have different attributes preserved.  For that you'd have to add a flag value to AXL_Add() or something.  Like: `AXL_Add(id, src, dest, AXL_PRESERVE_XADDR | AXL_PRESERVE_PERMISSIONS)`.

I would vote in favour of just setting it globally in `AXL_Config()`, since the common case will be that they'd want the same attributes preserved for all their files.

PR #68, opened 2020-06-07
copy uid, gid, permissions, and timestamps

adammoody:
If AXL_COPY_METADATA=1, then copy file metadata, including uid, gid, permissions, and timestamps.

This only applies to pthread and sync so far, with the assumption that IBM BB and Cray Datawarp may already be copying this metadata.  If not, it can be added to those.

Resolves: https://github.com/ECP-VeloC/AXL/issues/41

### DIFF
--- a/src/axl_io.c
+++ b/src/axl_io.c
@@ -17,6 +17,10 @@
 #include <errno.h>
 #include <string.h>
 
+#include <unistd.h>
+#include <sys/types.h>
+#include <stdint.h>
+
 #include "axl_internal.h"
 
 /* Configurations */
@@ -27,6 +31,16 @@
 #ifndef AXL_OPEN_USLEEP
 #define AXL_OPEN_USLEEP (100)
 #endif
+
+/* TODO: ugly hack until we get a configure test */
+#if defined(__APPLE__)
+#define HAVE_STRUCT_STAT_ST_MTIMESPEC_TV_NSEC 1
+#else
+#define HAVE_STRUCT_STAT_ST_MTIM_TV_NSEC 1
+#endif
+// HAVE_STRUCT_STAT_ST_MTIME_N
+// HAVE_STRUCT_STAT_ST_UMTIME
+// HAVE_STRUCT_STAT_ST_MTIME_USEC
 
 /* returns user's current mode as determine by their umask */
 mode_t axl_getmode(int read, int write, int execute) {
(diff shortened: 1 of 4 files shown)


src/axl_io.c
```diff
@@ -408,8 +606,17 @@ int axl_file_copy(const char* src_file, const char* dst_file, unsigned long buf_
...
 
-    /* unlink the file if the copy failed */
-    if (rc != AXL_SUCCESS) {
+    if (rc == AXL_SUCCESS) {
+        /* succeeded in copying the file, now copy uid/gid,
+         * permissions, and timestamps */
+        if (axl_copy_metadata) {
+            kvtree* meta = kvtree_new();
+            axl_meta_encode(src_file, meta);
+            axl_meta_apply(dst_file, meta);
```
tonyhutter, line 615: Do you really need to encode the values into a KVTree?  Why not just pass `sb` and `nsecs` to `axl_meta_apply()`?

adammoody: The main reason is code reuse.  These same meta_encode/meta_apply functions are also in the shuffile and redset components, so they are duplicated across three components now.  I'm hoping we can extract this logic into a new I/O component as future work, and then we can replace all of these calls to use the new common I/O component.

adammoody: Oh, and forgot to add, they are encoded to kvtrees in those other components because the info has to be saved to files and transferred across nodes with MPI.  The kvtrees make that easy to do.

tonyhutter:
Would you mind including this patch as well?  It add flags to `axl_cp` to preserve timestamps/permissions.
```diff
diff --git a/test/axl_cp.c b/test/axl_cp.c
index 44d1516..62e0a11 100644
--- a/test/axl_cp.c
+++ b/test/axl_cp.c
@@ -6,9 +6,11 @@
 #include <errno.h>
 #include <sys/stat.h>
 #include <signal.h>
+#include <stdlib.h>
 #include "axl.h"
 
 int id = -1;
+char *old_env = NULL;
 
 /* Return 1 if path is a directory */
 static int
@@ -51,9 +53,11 @@ axl_xfer_str_to_xfer(const char *xfer_str)
 static void
 usage(void)
 {
-    printf("Usage: axl_cp [-r|-R] [-X xfer_type] SOURCE DEST\n");
-    printf("       axl_cp [-r|-R] [-X xfer_type] SOURCE... DIRECTORY\n");
+    printf("Usage: axl_cp [-ap] [-r|-R] [-X xfer_type] SOURCE DEST\n");
+    printf("       axl_cp [-ap] [-r|-R] [-X xfer_type] SOURCE... DIRECTORY\n");
     printf("\n");
+    printf("-a:             Archive mode.  Preserve permissions + times + recursive.  Implies -pr\n");
+    printf("-p:             Preserve permissions + times.\n");
     printf("-r|-R:          Copy directories recursively\n");
     printf("-X xfer_type:   AXL transfer type:  default native pthread sync dw bbapi cppr\n");
     printf("\n");
@@ -86,6 +90,9 @@ void sig_func(int signum)
         exit(rc);
     }
 
+    if (old_env)
+        setenv("AXL_COPY_METADATA", old_env, 1);
+
     exit(AXL_SUCCESS);
 }
 
@@ -107,9 +114,17 @@ main(int argc, char **argv) {
     action.sa_handler = sig_func;
     sigaction(SIGTERM, &action, NULL);
     char *state_file = NULL;
+    int preserve = 0;
 
-    while ((opt = getopt(argc, argv, "rRSX:")) != -1) {
+    while ((opt = getopt(argc, argv, "aprRSX:")) != -1) {
         switch (opt) {
+            case 'p':
+                preserve = 1;
+                break;
+            case 'a':
+                preserve = 1;
+                recursive = 1;
+                break;
             case 'X':
                 xfer_str = optarg;
                 break;
@@ -125,6 +140,10 @@ main(int argc, char **argv) {
                 exit(1);
         }
     }
+
+    if (preserve)
+       old_env = getenv("AXL_COPY_METADATA");
+
     args_left = argc - optind;
 
     /* Did they specify source(s) and destination? */
@@ -163,6 +182,9 @@ main(int argc, char **argv) {
         xfer = AXL_XFER_SYNC;
     }
 
+    if (preserve)
+        setenv("AXL_COPY_METADATA", "1", 1);
+
     rc = AXL_Init(state_file);
     if (rc != AXL_SUCCESS) {
         printf("AXL_Init() failed (error %d)\n", rc);
@@ -212,5 +234,8 @@ main(int argc, char **argv) {
         return rc;
     }
 
+    if (old_env)
+        setenv("AXL_COPY_METADATA", old_env, 1);
+
     return AXL_SUCCESS;
 }
```

adammoody:
@tonyhutter , yeah that's good.  Thanks.  Go ahead and directly push that commit to the existing metadata branch.

tonyhutter:
Thanks, pushed.

adammoody:
Thanks for adding your patch, @tonyhutter !

Issue #66, opened 2020-05-28
Use different filename while file is being transferred


tonyhutter:
We need a way for AXL to signify to the user that a file transfer is complete, beyond just calling `AXL_Wait()`.  Reasons are discussed here and in subsequent comments:
https://github.com/LLNL/scr/issues/163#issuecomment-635018484

The easiest solution would be tack on an `._AXL` extension to the filename while it's being transferred:
```
ckpt_0._AXL
```
That way, it gives a hint to the user that "my `ckpt_0` file is still being transferred by AXL" and they could `ls -l` it to track progress.  After the transfer is complete AXL would move the file to it's final name.

PR #73, opened 2020-07-07
Use temporary file name when transferring

tonyhutter:
This patch will add an temporary AXL extension to the file while it is copying.  The extension is removed after the file copy
is complete.  For example, if you're copying to `/tmp/file1`, AXL will actually copy to `/tmp/file1._AXL`, and then rename to
`/tmp/file1` at the end of the transfer. 

Additionally, a BB API transfer will also encode the transfer handle number into the extension like:

`/tmp/file1._AXL13456`

That way we can later recover the transfer handle number.

Fixes: #66

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -44,6 +44,9 @@ typedef enum {
     AXL_XFER_STATE_CANCELED,   /* transfer was AXL_Cancel'd */
 } axl_xfer_state_t;
 
+/* Temporary extension added onto files while they're being transferred. */
+#define AXL_EXTENSION "._AXL"
+
 /*
 =========================================
 Global Variables
@@ -332,6 +335,23 @@ static int path_type(const char *path) {
 }
 
 /*
+ * Given a path name to a file ('path'), add on an extra AXL temporary extension
+ * in the form of: path._AXL[extra].  So if path = '/tmp/file1' and
+ * extra = "1234", then the new name is /tmp/file1._AXL1234.  The new name is
+ * allocated and returned as a new string (which must be freed).
+ *
+ * 'extra' can be optionally used to store additional information about the
+ * transfer, such as the BB API transfer handle number, or it can be left NULL.
+ */
+static char * axl_add_extension(const char *path, const char *extra)
+{
+    char *tmp = NULL;
+    asprintf(&tmp, "%s%s%s", path, AXL_EXTENSION, extra ? extra : "");
+    return tmp;
+}
+
+
+/*
  * Add a file to an existing transfer handle.  No directories.
  *
  * If the file's destination path doesn't exist, then automatically create the
@@ -341,6 +361,8 @@ static int
 __AXL_Add (int id, const char* src, const char* dest)
 {
     kvtree* file_list = NULL;
+    char *newdest = NULL;
+    char *extra = NULL;
 
     axl_xfer_t xtype = AXL_XFER_NULL;
     axl_xfer_state_t xstate = AXL_XFER_STATE_NULL;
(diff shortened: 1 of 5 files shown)


src/axl.c
```diff
@@ -610,6 +657,74 @@ int AXL_Dispatch (int id)
...
 
+/*
+ * Given a path with an AXL temporary extension, allocate an return a new
+ * string with the extension removed.  Also, if extra is specified, return
+ * a pointer to the offset in 'path_with_extension' where the 'extra' field
+ * is.
+ *
+ * Returns the new allocated path string on success, or NULL on error.
+ */
+static char * axl_remove_extension(char *path_with_extension, char **extra)
```
adammoody, line 668: Does this need to account for the IBM handle if one is attached?

src/axl.c
```diff
@@ -610,6 +657,74 @@ int AXL_Dispatch (int id)
...
+{
+	int i;
+	size_t size = strlen(path_with_extension);
+    char *ext = AXL_EXTENSION;
+    size_t ext_len = sizeof(AXL_EXTENSION) - 1; /* -1 for '\0' */
```
adammoody, line 673: Just missing some spaces here I think
```suggestion
      char *ext = AXL_EXTENSION;
      size_t ext_len = sizeof(AXL_EXTENSION) - 1; /* -1 for '\0' */
```

tonyhutter: Fixed in latest push

src/axl.c
tonyhutter: Correct, it accounts for the transfer handle

adammoody: Oh, I see.  My first impression was that extra was an input parameter rather than an output value.

adammoody:
Looks good to me.  Thanks @tonyhutter .

adammoody:
@tonyhutter , would you please update this PR to use 4-space indents and add braces even to single-line if/while blocks throughout?

adammoody:
Can you remind me, what is the plan to lookup the BB transfer ID for a cancel?

I thought I had that straight, but now I'm forgetting again.  The AXL user will call either ```AXL_Cancel(id)``` or ```AXL_Stop()```.

tonyhutter:
Sorry about the indents - I've fixed those in my latest push.

Correct, it's `AXL_Stop()` or `AXL_Cancel(id)` in the normal case.  For old transfers, the plan is to add a new function to AXL to deal to cleaning them up.  We would either update `AXL_Finalize()` -> `AXL_Finalize(char *path)` or add a new `AXL_Cleanup(char *path)`, where `path` is an individual file or a directory of files to cleanup.  That function would remove old transfer files with the ._AXL extension, or for BBAPI, it would check the transfer status for the file, and rename the file to its final name or delete it depending on the BB transfer status (success or failure).  This is to be done in a follow-on PR.

adammoody:
What do you mean by "normal case" vs "old transfers"?

tonyhutter:
normal case - your application calls `AXL_Cancel(id)` to purposefully cancel an ongoing transfer that it started.
old transfers - your application starts a transfer, the app crashes and restarts, and now your app is dealing with your old transfer files (the ones with the `._AXL` extensions)

adammoody:
Ok, yes, we'll need to think through our use cases with the cleanup API.

On a related topic, the reason the AXL state file exists is so that one can call Cancel and Stop on transfers that were initiated before the process that is making the Cancel/Stop call.  AXL should record whatever it needs in this file to maintain its state through process failure.

The plan was for SCR's integration to look like this:
```
<run 1 starts>
AXL_Init(/path/to/state/file)
<define AXL transfer>
<AXL records info about that transfer in its state file>
<start AXL transfer>
<run 1 dies>

<run 2 starts>
AXL_Init(/path/to/state/file)
<AXL reads in its state file, which contains info about all transfer that might be ongoing>
AXL_Stop()
```
The state file can record the mapping of AXL transfer id to things like the list of source and destination files associated with that id, the metadata for each of those files, the transfer method being used, and any vendor-specific information like the BB transfer handle.  With this, the application can use any or all of Test/Wait/Cancel/Stop after being restarted, but to do so, the application is required to provide a path to a file where AXL can store its state.

tonyhutter:
> With this, the application can use any or all of Test/Wait/Cancel/Stop after being restarted, but to do so, the application is required to provide a path to a file where AXL can store its state.

Sure, you could do an`AXL_Stop()`, but all the other functions require an AXL ID.  You'd need to add something to the AXL API for the user to recover those IDs.

Alternatively, if they pass a state_file in `AXL_Init()`, you could simply do:

1. For non-BB_API transfers, remove all the old ._AXL files in the transfer list.
2. For BB API transfers, check the BB API transfer status.
	- If it reports a successful transfer, rename the file to it's final filename, since it must have transferred in the background between jobs.  
	- If it reports the transfer is still in progress, cancel the transfer and delete the ._AXL file.
	- If it reports any other status, delete the ._AXL file.

adammoody:
That's right.  With the current AXL API, the caller of AXL is responsible for saving the AXL ID values somewhere if it wants to be able to use them across process failure.

adammoody:
Right now, SCR just calls AXL_Stop on a restart on every restart.  But that's overkill.

From past experience, we found that a significant number of failures just kill the application and otherwise leave all compute nodes intact (30%).  For those, the restarted job lands on the same nodes, all files exist, and there is no need to move files around.  In that case, there is no need to actually stop the ongoing transfers.

A nice optimization that would could add to SCR is to allow the transfer to continue/restart whenever SCR identifies that all source files are still valid, which changes the above to this.
```
<run 2 starts>
AXL_Init(/path/to/state/file)
<AXL reads in its state file, which contains info about all transfer that might be ongoing>

SCR determines whether any source file must be rebuilt or shuffled
if so
  AXL_Stop() // stop all transfers
else
  AXL_Wait(id) // just let the previous transfer finish
```
Perhaps we want an ```AXL_Restart(id)``` or something to make things work better in the general case like pthreads or AXL could just kick off the new pthread transfer when it sees an existing, unfinished pthread transfer in its state file during ```AXL_Init```.

adammoody:
I think we have like 80% of what we need to support this test/wait/cancel/stop after a restart already working in AXL, but the requirements for that last 20% are not well defined yet.  We'll need to work from the SCR side for a while to figure out where the gaps are.  I can pitch in on the SCR side by later next week.

tonyhutter:
I opened #76 for continued discussion on the state_file.

Issue #74, opened 2020-07-16
Support optional AXL configuration that deletes source files when transfer completes


adammoody:
One potential future AXL user has requested the ability to have AXL automatically delete a source file after it has successfully transferred the file to its destination.  This feature should be optional, since some users want to also keep the source files around.

An open question is how to handle the case where a transfer moves more than one file, and a subset of those succeed and others error out.  Do we require an all or nothing when considering whether to delete the source files?

tonyhutter:
I think should copy the behaviour of `mv`.  I did a little experiment where I tried to move four 3MB files (`testfile[1-4]`) into a 10MB mount to see what would happen.  I got an ENOSPC while copying the 3rd file.  Here's what files I saw afterwards:

Source:
```
-rw-r----- 1 hutter2 hutter2 3145728 Jul 16 08:52 testfile3
-rw-r----- 1 hutter2 hutter2 3145728 Jul 16 08:52 testfile4
```

Dest:
```
-rw-r----- 1 hutter2 hutter2 3145728 Jul 16 08:52 testfile1
-rw-r----- 1 hutter2 hutter2 3145728 Jul 16 08:52 testfile2
-rw------- 1 root    root    2420736 Jul 16 08:52 testfile3
-rw------- 1 root    root          0 Jul 16 08:52 testfile4
```
So it looks like `mv` doesn't remove the source file, nor set the permissions on the destination file, until the file is fully transferred.

bnicolae:
Hmm, "mv" should not have permission to run as root. What you probably are observing is the behavior of the file system, which is something beyond the control of mv. I think the most appropriate approach for AXL would be to emulate how web browsers are downloading files: create a temp name on the destination mount point, try to complete the transfer, if successful rename the file and remove the source. On failure, nobody will work with incomplete files at any point, a subsequent transfer of the same file can detect partial progress and resume (if temp names are not random) and garbage collection is also easy

tonyhutter:
@bnicolae just for some context - the source files were owned by hutter2, but I was moving them to an ext4 partition that was owned by root (specifically, a file that I formatted ext4 and mounted loopback).  So I just did a `sudo mv ...` because I was too lazy to change the permissions of the mount before copying.

tonyhutter:
@bnicolae forgot to mention that as of https://github.com/ECP-VeloC/AXL/commit/d738fb7a973367f57d9480cba9b2b63717920374 AXL will use a temp file name while transferring.

PR #77, opened 2020-07-30
Add resumeable transfers, update state_file

tonyhutter:
- Change the API so that `AXL_Create()` takes in a state_file argument, and remove state_file from `AXL_Init()`.  Use macro mangling for backwards compatibility so that this doesn't break applications that call the functions the old way.  We can get rid of the macro mangling once the FILO/VeloC are updated.

- Add `AXL_Resume(id)` to resume a transfer.  This works for sync, pthreads, and BBAPI.  For BBAPI, there's no way to resume  a transfer, so we just keep the existing transfer going.

- Update `axl_cp` and test cases with all these changes.

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -2,6 +2,7 @@
 #include <string.h>
 #include <errno.h>
 #include <stdio.h>
+#include <pthread.h>
 
 /* PATH_MAX */
 #include <limits.h>
@@ -57,14 +58,85 @@ Global Variables
  * before transferring files */
 int axl_make_directories;
 
-/* AXL's flush file, SCR has one as well */
-char* axl_flush_file = NULL;
+/*
+ * Array for all the AXL_Create'd kvtree pointers.  It's indexed by the AXL id.
+ *
+ * Note: We only expand this array, we never shrink it.  This is fine since
+ * the user is only going to be calling AXL_Create() a handful of times.  It
+ * also simplifies the code if we never shrink it, and the extra memory usage
+ * is negligible, if any at all.
+ */
+kvtree** axl_kvtrees;
+static unsigned int axl_kvtrees_count = 0;
 
-/* Transfer handle unique IDs */
-static int axl_next_handle_UID = -1;
+static pthread_mutex_t id_lock = PTHREAD_MUTEX_INITIALIZER;
 
-/* tracks list of files written with transfer */
-kvtree* axl_file_lists = NULL;
+/*
+ * Allocate a new kvtree and return the AXL ID for it.  If state_file is
+ * specified, then populate the kvtree with it's data.
+ */
+static int axl_alloc_id(const char* state_file)
+{
+    kvtree* new;
+
+    new = kvtree_new();
+    if (state_file) {
+         /* initialize kvtree values from state_file if we have one */
+        if (access(state_file, F_OK) == 0 &&
+            kvtree_read_file(state_file, new) != KVTREE_SUCCESS) {
+            AXL_ERR("Couldn't read state file correctly");
+            return -1;
+        }
+
+        kvtree_util_set_str(new, AXL_KEY_STATE_FILE, state_file);
+    }
+
+    pthread_mutex_lock(&id_lock);
+    axl_kvtrees = realloc(axl_kvtrees, sizeof(struct kvtree *) * (axl_kvtrees_count + 1));
+    axl_kvtrees[axl_kvtrees_count] = new;
+    axl_kvtrees_count++;
+    pthread_mutex_unlock(&id_lock);
+    return axl_kvtrees_count - 1;
+}
+
+/* Remove the state file for an id, if one exists */
+static void axl_remove_state_file(int id)
+{
+    kvtree* file_list = axl_kvtrees[id];
+    char* state_file = NULL;
+
+    if (kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE,
+        &state_file) == KVTREE_SUCCESS) {
+        axl_file_unlink(state_file);
+    }
+}
+
+/* Remove the ID, we don't need it anymore */
+static void axl_free_id(int id)
+{
+    axl_remove_state_file(id);
+
+    pthread_mutex_lock(&id_lock);
+
+    /* kvtree_delete() will set axl_kvtrees[id] = NULL */
+    kvtree_delete(&axl_kvtrees[id]);
+
+    pthread_mutex_unlock(&id_lock);
+}
+
+/*
+ * If the user specified a state_file then write our kvtree to it. If not, then
+ * do nothing.
+ */
+static void axl_write_state_file(int id)
+{
+    kvtree* file_list = axl_kvtrees[id];
+    char* state_file = NULL;
+    if (kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE,
+        &state_file) == KVTREE_SUCCESS) {
+        kvtree_write_file(state_file, file_list);
+    }
+}
 
 /* track whether file metadata should also be copied */
 int axl_copy_metadata = 0;
@@ -79,7 +151,7 @@ static int axl_get_info(int id, kvtree** list, axl_xfer_t* type, axl_xfer_state_
     *state = AXL_XFER_STATE_NULL;
 
     /* lookup transfer info for the given id */
-    kvtree* file_list = kvtree_get_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
+    kvtree* file_list = axl_kvtrees[id];
     if (file_list == NULL) {
         AXL_ERR("Could not find fileset for UID %d", id);
         return AXL_FAILURE;
@@ -162,7 +234,7 @@ API Functions
 
 /* Read configuration from non-AXL-specific file
   Also, start up vendor specific services */
-int AXL_Init (const char* state_file)
+int __AXL_Init (const char* state_file)
 {
     int rc = AXL_SUCCESS;
 
@@ -191,27 +263,9 @@ int AXL_Init (const char* state_file)
         axl_make_directories = atoi(val);
     }
 
-    /* make a copy of the path to file to record our state */
     if (state_file != NULL) {
-        axl_flush_file = strdup(state_file);
-    }
-
-    axl_next_handle_UID = 0;
-    axl_file_lists = kvtree_new();
-
-    /* initialize values from state file if we have one */
-    if (axl_flush_file != NULL) {
-        /* initialize axl_file_lists from file if we have one */
-        kvtree_read_file(axl_flush_file, axl_file_lists);
-
-        /* initialize handle id using highest id in file */
-        kvtree* ids = kvtree_get(axl_file_lists, AXL_KEY_HANDLE_UID);
-        kvtree_sort_int(ids, KVTREE_SORT_DESCENDING);
-        kvtree_elem* elem = kvtree_elem_first(ids);
-        if (elem != NULL) {
-            int id = kvtree_elem_key_int(elem);
-            axl_next_handle_UID = id;
-        }
+        printf("WARNING: Passing a state_file to AXL_Init() is deprecated." \
+                "Pass it to AXL_Create(id, type, state_file) instead.\n");
     }
 
 #ifdef HAVE_BBAPI
@@ -244,12 +298,6 @@ int AXL_Finalize (void)
     }
 #endif
 
-    /* delete flush file if we have one */
-    if (axl_flush_file != NULL) {
-        axl_file_unlink(axl_flush_file);
-    }
-    axl_free(&axl_flush_file);
-
     return rc;
 }
 
@@ -257,15 +305,26 @@ int AXL_Finalize (void)
  * Type specifies a particular method to use
  * Name is a user/application provided string
  * Returns an ID to the transfer handle */
-int AXL_Create (axl_xfer_t xtype, const char* name)
+int __AXL_Create(axl_xfer_t xtype, const char* name, const char *state_file)
 {
-    /* Generate next unique ID */
-    int id = ++axl_next_handle_UID;
+    int id;
+    int reload_from_state_file = 1;
+    kvtree* file_list = NULL;
+    axl_xfer_t old_xtype = AXL_XFER_NULL;
+    axl_xfer_state_t old_xstate = AXL_XFER_STATE_NULL;
 
-    if (xtype == AXL_XFER_DEFAULT) {
-        xtype = axl_detect_default_xfer();
-    } else if (xtype == AXL_XFER_NATIVE) {
-        xtype = axl_detect_native_xfer();
+    /* Generate next unique ID */
+    id = axl_alloc_id(state_file);
+    if (id < 0) {
+        return id;
+    }
+
+   /*
+     * If no state file is passed, or it doesn't exist yet, then we're starting
+     * from scratch.
+     */
+    if (!state_file || (state_file && access(state_file, F_OK) != 0)) {
+        reload_from_state_file = 0;
     }
 
     /* Create an entry for this transfer handle
@@ -280,11 +339,33 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
      *       SOURCE
      *     STATE
      *       CREATED */
-    kvtree* file_list = kvtree_set_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
+    if (reload_from_state_file) {
+        if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
+            AXL_ERR("Couldn't get kvtree info");
+            return -1;
+        }
+
+        if (xtype != old_xtype) {
+            AXL_ERR("Transfer type %d doesn't match transfer type %d "
+                "in state_file", xtype, old_xtype);
+            return -1;
+        }
+    } else {
+        /* We're not loading from an existing state file */
+        if (xtype == AXL_XFER_DEFAULT) {
+            xtype = axl_detect_default_xfer();
+        } else if (xtype == AXL_XFER_NATIVE) {
+            xtype = axl_detect_native_xfer();
+        }
+
+        file_list = axl_kvtrees[id];
+        kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
+        kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
+        kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
+    }
+
+    /* Allow caller to assign a new name */
     kvtree_util_set_str(file_list, AXL_KEY_UNAME, name);
-    kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
-    kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
-    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
 
     /* create a structure based on transfer type */
     int rc = AXL_SUCCESS;
@@ -296,7 +377,9 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
     case AXL_XFER_PTHREAD:
         break;
     case AXL_XFER_ASYNC_BBAPI:
-        rc = axl_async_create_bbapi(id);
+        if (!reload_from_state_file) {
+            rc = axl_async_create_bbapi(id);
+        }
         break;
     default:
         AXL_ERR("Unknown transfer type (%d)", (int) xtype);
@@ -306,13 +389,11 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
 
     /* clear entry from our list if something went wrong */
     if (rc != AXL_SUCCESS) {
-        kvtree_unset_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
+        axl_free_id(id);
         id = -1;
-    }
-
-    /* write data to file if we have one */
-    if (axl_flush_file) {
-        kvtree_write_file(axl_flush_file, axl_file_lists);
+    } else {
+        /* Success, write data to file if we have one */
+        axl_write_state_file(id);
     }
 
     return id;
@@ -444,9 +525,7 @@ __AXL_Add (int id, const char* src, const char* dest)
     }
 
     /* write data to file if we have one */
-    if (axl_flush_file) {
-        kvtree_write_file(axl_flush_file, axl_file_lists);
-    }
+    axl_write_state_file(id);
 
     return rc;
 }
@@ -559,8 +638,14 @@ int AXL_Add (int id, const char *src, const char *dest)
     return rc;
 }
 
-/* Initiate a transfer for all files in handle ID */
-int AXL_Dispatch (int id)
+/*
+ * Initiate a transfer for all files in handle ID.  If resume is set to 1, then
+ * attempt to resume the transfers from the existing destination files.
+ *
+ * If resume is set, but some of the destination files don't exist, then do
+ * a normal transfer for them.
+ */
+int __AXL_Dispatch (int id, int resume)
 {
     /* lookup transfer info for the given id */
     kvtree* file_list = NULL;
@@ -573,9 +658,29 @@ int AXL_Dispatch (int id)
         AXL_ERR("Could not find transfer info for UID %d", id);
         return AXL_FAILURE;
     }
+    if (resume) {
+        switch (xstate) {
+        case AXL_XFER_STATE_NULL:
+            AXL_ERR("Starting AXL_Resume() in state AXL_XFER_STATE_NULL");
+            return AXL_FAILURE;
+            break;
+        case AXL_XFER_STATE_COMPLETED:
+            /*
+             * We're trying to resume a transfer that's already completed, so
+             * we're done!
+             */
+            return AXL_SUCCESS;
+            break;
 
-    /* check that handle is in correct state to dispatch */
-    if (xstate != AXL_XFER_STATE_CREATED) {
+        case AXL_XFER_STATE_CREATED:
+        case AXL_XFER_STATE_DISPATCHED:
+        case AXL_XFER_STATE_WAITING:
+        case AXL_XFER_STATE_CANCELED:
+            /* Start or resume the transfer */
+            break;
+        }
+    } else if (xstate != AXL_XFER_STATE_CREATED) {
+        /* check that handle is in correct state to dispatch */
         AXL_ERR("Invalid state to dispatch UID %d", id);
         return AXL_FAILURE;
     }
@@ -605,7 +710,7 @@ int AXL_Dispatch (int id)
              kvtree_util_set_int(file_list, AXL_BBAPI_KEY_FALLBACK, 1);
         }
 
-        if (!axl_bbapi_in_fallback(id)) {
+        if (!axl_bbapi_in_fallback(id) && !resume) {
             char *src = NULL;
             int rc;
 
@@ -629,16 +734,33 @@ int AXL_Dispatch (int id)
     int rc = AXL_SUCCESS;
     switch (xtype) {
     case AXL_XFER_SYNC:
-        rc = axl_sync_start(id);
+        if (resume) {
+            rc = axl_sync_resume(id);
+        } else {
+            rc = axl_sync_start(id);
+        }
         break;
     case AXL_XFER_PTHREAD:
-        rc = axl_pthread_start(id);
+        if (resume) {
+            rc = axl_pthread_resume(id);
+        } else {
+            rc = axl_pthread_start(id);
+        }
         break;
     case AXL_XFER_ASYNC_DW:
+        if (resume) {
+            AXL_ERR("AXL_Resume() isn't supported yet for DW");
+            rc = AXL_FAILURE;
+            break;
+        }
         rc = axl_async_start_datawarp(id);
         break;
     case AXL_XFER_ASYNC_BBAPI:
-        rc = axl_async_start_bbapi(id);
+        if (resume) {
+            rc = axl_async_resume_bbapi(id);
+        } else {
+            rc = axl_async_start_bbapi(id);
+        }
         break;
     /* case AXL_XFER_ASYNC_CPPR:
         rc = axl_async_start_cppr(id); */
@@ -650,11 +772,19 @@ int AXL_Dispatch (int id)
     }
 
     /* write data to file if we have one */
-    if (axl_flush_file) {
-        kvtree_write_file(axl_flush_file, axl_file_lists);
-    }
+    axl_write_state_file(id);
 
     return rc;
+}
+
+int AXL_Dispatch(int id)
+{
+    return __AXL_Dispatch(id, 0);
+}
+
+int AXL_Resume(int id)
+{
+    return __AXL_Dispatch(id, 1);
 }
 
 /*
@@ -861,9 +991,7 @@ end:
     kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_COMPLETED);
 
     /* write data to file if we have one */
-    if (axl_flush_file) {
-        kvtree_write_file(axl_flush_file, axl_file_lists);
-    }
+    axl_write_state_file(id);
 
     return rc;
 }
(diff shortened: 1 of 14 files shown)


tonyhutter:
This is still WIP while I'm testing the BB API transfer resume.

tonyhutter:
I've substantially revamped this code, including adding BB API test cases.  It should be ready to review. I'm removing the WIP tag.

src/axl.h
```diff
@@ -45,26 +45,59 @@ typedef enum {
...
+ * Returns an AXL ID, or negative number on error.
+ *
+ * NOTE: AXL_Create() used to only take in xtype and name.  The state_file arg
+ * was added later.  We do some macro mangling to make sure both 2-arg and
+ * 3-arg versions of AXL_Create() work, but the 2-arg version is deprecated
+ * and should not be used.
+ */
+#define AXL_Create(type, name, ...) \
+        __AXL_Create(type, name, GET_ARG0(__VA_ARGS__))
+int __AXL_Create(axl_xfer_t xtype, const char* name, const char *state_file);
```
adammoody, line 82: We need to define whether it's an error if someone names a state file, which exists and is read in, but then they have provide an xtype and/or name here that do not match the values recorded in the file.

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
 
-/* Transfer handle unique IDs */
-static int axl_next_handle_UID = -1;
+/* Remove the ID, we don't need it anymore */
+static void axl_free_id(int id)
+{
+    axl_remove_state_file(id);
+    pthread_mutex_lock(&id_lock);
+    kvtree_delete(&axl_kvtrees[id]);
+    axl_kvtrees[id] = NULL;
```
adammoody, line 121: This line can be dropped, as kvtree_delete already sets this to NULL
```suggestion
```

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
+}
+
+/* Remove the state file for an id, if one exists */
+static void axl_remove_state_file(int id)
+{
+    kvtree *file_list = axl_kvtrees[id];
+    char *state_file = NULL;
+    kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE, &state_file);
+    if (state_file && strcmp(state_file, "") != 0) {
+        unlink(state_file);
```
adammoody, line 111: Let's use the axl_file_unlink call here.

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
+ * specified, then populate the kvtree with it's data.
+ */
+static int axl_alloc_id(const char* state_file)
+{
+    kvtree *new;
+
+    new = kvtree_new();
+    if (state_file) {
+         /* initialize kvtree values from state_file if we have one */
+        kvtree_read_file(state_file, new);
```
adammoody, line 85: It's possible we could be reading a corrupted file here, e.g., node died before kvtree file was sync'd.  We should check and return an error here if that happens.

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
+    } else {
+        /*
+         * No state file, just put in an empty string since you can't use
+         * a NULL for a string in kvtree.
+         */
+        state_file = "";
+    }
+
+    kvtree_util_set_str(new, AXL_KEY_STATE_FILE, state_file);
```
adammoody, line 94: It would be cleaner to avoid setting this field at all when the user passes state_file == NULL.  We can later check whether it was set with ``kvtree_util_get_str(kvtree, AXL_KEY_STATE_FILE, &state_file) == KV_TREE_SUCCESS)``.

src/axl.c
```diff
@@ -984,41 +1066,36 @@ int AXL_Stop ()
...
-
     /* cancel each active id */
-    int i;
-    for (i = 0; i < numids; i++) {
-        int id = ids[i];
+    int id;
+    for (id = 0; id < axl_kvtrees_count; id++) {
+        if (!axl_kvtrees[id])
+            continue;
+
```
adammoody, line 1074: To keep with the style, please add braces around all code blocks.

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
 
-/* tracks list of files written with transfer */
-kvtree* axl_file_lists = NULL;
+/*
+ * If the user specified a state_file then write our kvtree to it. If not, then
+ * do nothing.
+ */
+static void axl_write_state_file(int id)
+{
+    kvtree *file_list = axl_kvtrees[id];
```
adammoody, line 131: To match existing style, please move the stars next to the type rather than the variable name.
```suggestion
    kvtree* file_list = axl_kvtrees[id];
```

src/axl.c
```diff
@@ -57,14 +58,83 @@ Global Variables
...
+    char *state_file = NULL;
+    kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE, &state_file);
+    if (state_file && strcmp(state_file, "") != 0) {
```
adammoody, line 134:
```suggestion
    if (kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE, &state_file) == KVTREE_SUCCES) {
```

src/axl.c
```diff
@@ -970,12 +1054,10 @@ int AXL_Free (int id)
...
 
     /* forget anything we know about this id */
-    kvtree_unset_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
+    axl_free_id(id);
 
     /* write data to file if we have one */
-    if (axl_flush_file) {
-        kvtree_write_file(axl_flush_file, axl_file_lists);
-    }
+    axl_write_state_file(id);
```
adammoody, line 1060: Guessing you don't want to write the file out right after you've deleted it.

src/axl_async_bbapi.c
```diff
@@ -279,6 +280,7 @@ int axl_async_create_bbapi(int id) {
         return bb_check(rc);
     }
 
+
```
adammoody, line 283:
```suggestion
```

src/axl_async_bbapi.c
```diff
@@ -331,22 +333,36 @@ int axl_async_add_bbapi (int id, const char* source, const char* dest) {
...
+        if (old_status == AXL_STATUS_INPROG) {
+            /* Our transfers are already going.  Nothing to do */
+            return AXL_SUCCESS;
+        }
+    }
 
     /* mark this transfer as in progress */
     kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_INPROG);
 
+
```
adammoody, line 365:
```suggestion
```

src/axl_io.c
```diff
@@ -558,6 +590,21 @@ int axl_file_copy(const char* src_file, const char* dst_file, unsigned long buf_
         *crc = crc32(0L, Z_NULL, 0);
     }
 
+    /* Resume the transfer to our destination file where we left off */
+    if (resume) {
```
adammoody, line 594: I don't think the crc32 value won't be computed correctly when resuming a file.  I don't think we're using crcs in AXL, so one option is to consider dropping it?  I guess the question is whether we think users would like AXL to compute crc's as an option.

src/axl_io.c
```diff
@@ -593,6 +640,12 @@ int axl_file_copy(const char* src_file, const char* dst_file, unsigned long buf_
             copying = 0;
             rc = AXL_FAILURE;
         }
+        total_copied += nread;
+
+        /* Possibly pause our transfer for unit tests */
+        while (total_copied >= pause_after) {
+            pause_after = axl_debug_pause_after();
+        }
```
adammoody, line 648: This code is a bit convoluted when thinking about what it does.  Also, it's a bit strange to call this function repeatedly even when the option is not enabled.  Was the goal of writing it this way to make it easier to break the look in a debugger or for some other reason?

src/axl.c
```diff
@@ -573,9 +637,8 @@ int AXL_Dispatch (int id)
         AXL_ERR("Could not find transfer info for UID %d", id);
         return AXL_FAILURE;
     }
-
     /* check that handle is in correct state to dispatch */
-    if (xstate != AXL_XFER_STATE_CREATED) {
+    if (xstate != AXL_XFER_STATE_CREATED && !resume) {
```
adammoody, line 641: We dropped the check on the state when resuming.  I know resume needs to deal with both CREATED and DISPATCHED, but this might open things up for the user to call in with other states like COMPLETED.  Do the methods for all of the various backends work for all possible states?

adammoody:
Thanks, @tonyhutter .  I've taken a first pass.  I feel like this is complicated enough I'll want another pass or two as we work through it.

src/axl_io.c
adammoody: For example, I think the intent would be easier for someone to parse with something like:
```
if (pause_after != ULONG_MAX && total_copied >= pause_after) {
  while(1);
}
```

src/axl_io.c
adammoody: If users would like AXL to compute crcs, we'd want to keep it, but then code needs to be added to handle resume operations.

tonyhutter: If we're not using the CRC code, my vote would be to remove it in a separate PR and add it back in later if needed.

adammoody: Yeah, I agree we can remove it and add it back later.  Let's take those lines out in that case.

src/axl_io.c
tonyhutter: It's written the way it is so you can pause and unpause a transfer (by setting a different offset).  The cost is a `getenv()` on every megabyte read, which shouldn't be a big deal.  As you said though, it would be better to just check for the existence of `AXL_DEBUG_PAUSE_AFTER` at the beginning of the transfer, and if it's set, then continue to poll getenv() like it's doing now, since during test we're always going to be setting that envvar before starting the transfer.

adammoody: I don't think there is an easy way for someone to change the value that would be returned from ```getenv("AXL_DEBUG_PAUSE_AFTER")``` though, right?  How were you thinking someone would change what ```axl_debug_pause_after()``` is returning?

src/axl.c
tonyhutter: Good point. What do you think about these per-state actions if you call AXL_Resume():

AXL_XFER_STATE_NULL: return error
AXL_XFER_STATE_WAITING: Try to resume where you left off.
AXL_XFER_STATE_COMPLETED: Don't do anything.  Return success.
AXL_XFER_STATE_CANCELED: Try to resume where you left off.

tonyhutter:
@adammoody thanks for taking a first look.  If I didn't comment on something specificity, then assume I'll fix it in my next push.

src/axl_io.c
tonyhutter: `export AXL_DEBUG_PAUSE_AFTER=<new value>`

tonyhutter:
@adammoody I think I addressed all your comments in my latest push.

src/axl_io.c
tonyhutter: @adammoody you were right about this - if you export a envar after a program runs, the value is not picked up/changed.  So I'll need to update this.

tonyhutter: My latest push adds in the change you suggested

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
-char* axl_flush_file = NULL;
+/*
+ * Array for all the AXL_Create'd kvtree pointers.  It's indexed by the AXL id.
+ *
+ * Note: We only expand this array, we never shrink it.  This is fine since
+ * the user is only going to be calling AXL_Create() a handful of times.  It
+ * also simplifies the code if we never shrink it, and the extra memory usage
+ * is negligible, if any at all.
+ */
+kvtree **axl_kvtrees;
```
adammoody, line 69:
```suggestion
kvtree** axl_kvtrees;
```

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+    if (state_file) {
+         /* initialize kvtree values from state_file if we have one */
+        if (access(state_file, F_OK) == 0 &&
+            kvtree_read_file(state_file, new) != KVTREE_SUCCESS) {
+            AXL_ERR("Couldn't read state file correctly\n");
+            return AXL_FAILURE;
+        }
+
+        kvtree_util_set_str(new, AXL_KEY_STATE_FILE, state_file);
+
```
adammoody, line 92:
```suggestion
```

src/axl.c
```diff
@@ -244,28 +297,30 @@ int AXL_Finalize (void)
...
 {
+    int id;
+    int reload_from_state_file = 1;
+    kvtree* file_list = NULL;
+    axl_xfer_t old_xtype = AXL_XFER_NULL;
+    axl_xfer_state_t old_xstate = AXL_XFER_STATE_NULL;
+
     /* Generate next unique ID */
-    int id = ++axl_next_handle_UID;
+    id = axl_alloc_id(state_file);
```
adammoody, line 316: Looks like we need to catch and respond to the error here.

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+{
+    kvtree *new;
+
+    new = kvtree_new();
+    if (state_file) {
+         /* initialize kvtree values from state_file if we have one */
+        if (access(state_file, F_OK) == 0 &&
+            kvtree_read_file(state_file, new) != KVTREE_SUCCESS) {
+            AXL_ERR("Couldn't read state file correctly\n");
+            return AXL_FAILURE;
```
adammoody, line 88: Let's return -1 here, since we've been using that value to represent a bad id.  There is no guarantee that AXL_FAILURE will always be defined as -1.  If someone decides to change its definition to be a positive value instead like 1, this will show up as a rare bug they might overlook.

src/axl.c
```diff
@@ -280,11 +335,31 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
...
      *       CREATED */
-    kvtree* file_list = kvtree_set_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
-    kvtree_util_set_str(file_list, AXL_KEY_UNAME, name);
-    kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
-    kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
-    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
+    if (reload_from_state_file) {
+        if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
+            AXL_ERR("Couldn't get kvtree info\n");
+            return AXL_FAILURE;
```
adammoody, line 341: Let's return -1 instead of AXL_FAILURE through all of Create for the same reason.

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+    return axl_kvtrees_count - 1;
+}
+
+/* Remove the state file for an id, if one exists */
+static void axl_remove_state_file(int id)
+{
+    kvtree* file_list = axl_kvtrees[id];
+    char *state_file = NULL;
+    kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE, &state_file);
+    if (state_file && strcmp(state_file, "") != 0) {
```
adammoody, line 109:
```suggestion
    char* state_file = NULL;
    if (kvtree_util_get_str(file_list, AXL_KEY_STATE_FILE, &state_file) == KVTREE_SUCCESS) {
```

src/axl.c
```diff
@@ -280,11 +335,31 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
...
-    kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
-    kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
-    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
+    if (reload_from_state_file) {
+        if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
+            AXL_ERR("Couldn't get kvtree info\n");
+            return AXL_FAILURE;
+        }
+
+        if (xtype != old_xtype) {
```
adammoody, line 344: Let's also check that the name matches.

I haven't fully thought through things yet, but I feel like we'll have a race condition we need to figure out here when we get to that.

src/axl_io.c
```diff
@@ -494,10 +497,32 @@ int axl_meta_apply(const char* file, const kvtree* meta)
...
+ * been copied.  It's only used for test cases (like testing resuming a
+ * transfer).
+ *
+ * If the var is not set, return ULONG_MAX.
+ */
+static unsigned long axl_debug_pause_after(void)
+{
+    char *env= getenv("AXL_DEBUG_PAUSE_AFTER");
+    if (!env)
+        return ULONG_MAX;
```
adammoody, line 512:
```suggestion
    char* env= getenv("AXL_DEBUG_PAUSE_AFTER");
    if (!env) {
        return ULONG_MAX;
    }
```

src/axl_io.c
adammoody: The crc code is still here.

adammoody: Oh, I see you're suggesting to remove it in a later PR.  Sure, we can do that.

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
 
-/* tracks list of files written with transfer */
-kvtree* axl_file_lists = NULL;
+/*
+ * Allocate a new kvtree and return the AXL ID for it.  If state_file is
+ * specified, then populate the kvtree with it's data.
+ */
+static int axl_alloc_id(const char* state_file)
+{
+    kvtree *new;
```
adammoody, line 80:
```suggestion
    kvtree* new;
```

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+}
+
+/*
+ * If the user specified a state_file then write our kvtree to it. If not, then
+ * do nothing.
+ */
+static void axl_write_state_file(int id)
+{
+    kvtree* file_list = axl_kvtrees[id];
+    char *state_file = NULL;
```
adammoody, line 133:
```suggestion
    char* state_file = NULL;
```

src/axl.c
```diff
@@ -573,9 +652,29 @@ int AXL_Dispatch (int id)
         AXL_ERR("Could not find transfer info for UID %d", id);
         return AXL_FAILURE;
     }
+    if (resume) {
+        switch (xstate) {
+        case AXL_XFER_STATE_NULL:
+            AXL_ERR("Error: Starting AXL_Resume() in state AXL_XFER_STATE_NULL\n");
```
adammoody, line 658: I think our AXL_ERR macro tacks on a error prefix string and a newline, so you can leave those off.
```suggestion
            AXL_ERR("Starting AXL_Resume() in state AXL_XFER_STATE_NULL");
```

src/axl.c
```diff
@@ -280,11 +335,31 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
...
-    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
+    if (reload_from_state_file) {
+        if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
+            AXL_ERR("Couldn't get kvtree info\n");
+            return AXL_FAILURE;
+        }
+
+        if (xtype != old_xtype) {
+            AXL_ERR("Transfer type %d doesn't match transfer type %d "
+                "in state_file\n", xtype, old_xtype);
```
adammoody, line 346:
```suggestion
                "in state_file", xtype, old_xtype);
```

src/axl.c
```diff
@@ -280,11 +335,31 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
...
      *     STATE
      *       CREATED */
-    kvtree* file_list = kvtree_set_kv_int(axl_file_lists, AXL_KEY_HANDLE_UID, id);
-    kvtree_util_set_str(file_list, AXL_KEY_UNAME, name);
-    kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
-    kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
-    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
+    if (reload_from_state_file) {
+        if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
+            AXL_ERR("Couldn't get kvtree info\n");
```
adammoody, line 340:
```suggestion
            AXL_ERR("Couldn't get kvtree info");
```

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+static int axl_alloc_id(const char* state_file)
+{
+    kvtree *new;
+
+    new = kvtree_new();
+    if (state_file) {
+         /* initialize kvtree values from state_file if we have one */
+        if (access(state_file, F_OK) == 0 &&
+            kvtree_read_file(state_file, new) != KVTREE_SUCCESS) {
+            AXL_ERR("Couldn't read state file correctly\n");
```
adammoody, line 87:
```suggestion
            AXL_ERR("Couldn't read state file correctly");
```

src/axl.c
```diff
@@ -629,16 +728,33 @@ int AXL_Dispatch (int id)
...
-        rc = axl_pthread_start(id);
+        if (resume) {
+            rc = axl_pthread_resume(id);
+        } else {
+            rc = axl_pthread_start(id);
+        }
         break;
     case AXL_XFER_ASYNC_DW:
+        if (resume) {
+            AXL_ERR("AXL_Resume() isn't supported yet for DW\n");
```
adammoody, line 746:
```suggestion
            AXL_ERR("AXL_Resume() isn't supported yet for DW");
```

adammoody:
Thanks, @tonyhutter .  After those, I think we're on to the second phase of the resume review which is to consider failures at different points.  I was first thinking we would do that now, but it could be a lot of work, so maybe we could merge this (and the crc fix) then consider it in a third PR.  It's possible it may drive us to change how we have defined the AXL interface, though, so we should do it sooner rather than later if we go ahead and merge now.

Any snags are likely to show up after AXL completes a phase but the process dies just before AXL updates its state file to record the state change.  That means the resume will start up in an old state.  We need to think through each case in turn, meaing every AXL state x AXL transfer combination to make sure they all behave properly.

As an example, consider what happens if a process dies right here after it has renamed files but just before it marks the transfer as being complete:
https://github.com/ECP-VeloC/AXL/blob/d738fb7a973367f57d9480cba9b2b63717920374/src/axl.c#L864

A resumed pthread transfer would start up and see the transfer as DISPATCHED (not COMPLETED).  It also will not see the temp files, because they were in fact renamed.  So I think it will start over from scratch.  That's maybe inefficient, but I believe that might still work.  It will just end up writing all of the files as temp files again and then renaming again to clobber the final files.  I haven't looked through all of the code to make sure.  But that illustrates the kind of analysis we need to do to make sure resume works.

adammoody:
Consider things if we had already implemented the delete-after-complete feature: https://github.com/ECP-VeloC/AXL/issues/74

If we deleted the source files before writing COMPLETED to the state file, then the existing pthread resume would definitely no longer work.  It would try to restart, but it would be missing its source files.  We might need to add a new internal state to capture that distinction.  The interactions between everything get complicated and sort of a puzzle we need to think through.

src/axl_io.c
```diff
@@ -558,6 +590,21 @@ int axl_file_copy(const char* src_file, const char* dst_file, unsigned long buf_
...
+    /* Resume the transfer to our destination file where we left off */
+    if (resume) {
+        /* Seek to the end of our destination file, while recording offset */
+        start_offset = lseek(dst_fd, 0, SEEK_END);
+
+        /* Set the src file position to the start_offset of dst file */
+        if (lseek(src_fd, start_offset, SEEK_SET) != start_offset) {
+            AXL_ERR("Couldn't seek src file errno=%d %s",
+                    errno, strerror(errno));
+            axl_close(dst_file, dst_fd);
```
adammoody, line 602:
```suggestion
            axl_free(&buf);
            axl_close(dst_file, dst_fd);
```

adammoody:
We are also going to want a way for an application to remove a state file, for example to delete a corrupt state file or to get rid of a valid state file that the app cannot re-create because it lost track of the xfer type and/or name that was used in the original create call.

For now, we could say that it's the caller's responsibility to unlink the state file in those situations.

src/axl.c
tonyhutter: I was thinking they may want to set a new name, like if they include the job ID in the name ("simulation42").

tonyhutter:
@adammoody thanks for taking another look.  My latest push addresses the new fixes you suggested.

As far a failure cases go, I agree that we should document any ways we can think of where resuming the transfer can go wrong.  Then we can write test cases to exercise those scenarios to prove to ourselves that we have those bases covered.  I'm glad you brought up delete-after-transfer, because that is a place that could bite us if we're not being careful.

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
 
-/* Transfer handle unique IDs */
-static int axl_next_handle_UID = -1;
+/* Remove the ID, we don't need it anymore */
+static void axl_free_id(int id)
+{
+    pthread_mutex_lock(&id_lock);
+
+    /* kvtree_delete() will set axl_kvtrees[id] = NULL */
+    kvtree_delete(&axl_kvtrees[id]);
```
adammoody, line 120: is the tree being deleted here, and then a few lines later accessed in the call to axl_remove_state_file?

src/axl.c
```diff
@@ -57,14 +58,84 @@ Global Variables
...
+    axl_kvtrees_count++;
+    pthread_mutex_unlock(&id_lock);
+    return axl_kvtrees_count - 1;
+}
+
+/* Remove the state file for an id, if one exists */
+static void axl_remove_state_file(int id)
+{
+    kvtree* file_list = axl_kvtrees[id];
+    char *state_file = NULL;
```
adammoody, line 106:
```suggestion
    char* state_file = NULL;
```

src/axl.c
```diff
@@ -244,28 +297,32 @@ int AXL_Finalize (void)
...
+    int reload_from_state_file = 1;
+    kvtree* file_list = NULL;
+    axl_xfer_t old_xtype = AXL_XFER_NULL;
+    axl_xfer_state_t old_xstate = AXL_XFER_STATE_NULL;
+
     /* Generate next unique ID */
-    int id = ++axl_next_handle_UID;
+    id = axl_alloc_id(state_file);
+    if (id < 0)
+        return id;
```
adammoody, line 318:
```suggestion
    if (id < 0) {
        return id;
    }
```

src/axl_internal.h
```diff
@@ -6,17 +6,18 @@
...
 #define AXL_FAILURE (-1)
 
-/* AXL internal data structure
- * this structure is accessed by the each transfer interface
- * any changes of the existing structure must be documented */
-extern kvtree* axl_file_lists;
+/*
+ * A list of pointers to kvtrees, indexed by AXL ID.
+ */
+kvtree **axl_kvtrees;
```
adammoody, line 20:
```suggestion
kvtree** axl_kvtrees;
```

src/axl_io.c
```diff
@@ -494,10 +497,32 @@ int axl_meta_apply(const char* file, const kvtree* meta)
...
+ * been copied.  It's only used for test cases (like testing resuming a
+ * transfer).
+ *
+ * If the var is not set, return ULONG_MAX.
+ */
+static unsigned long axl_debug_pause_after(void)
+{
+    char* env= getenv("AXL_DEBUG_PAUSE_AFTER");
+    if (!env)
+        return ULONG_MAX;
```
adammoody, line 512:
```suggestion
    char* env = getenv("AXL_DEBUG_PAUSE_AFTER");
    if (!env) {
        return ULONG_MAX;
    }
```

adammoody:
> 
> 
> @adammoody thanks for taking another look. My latest push addresses the new fixes you suggested.
> 
> As far a failure cases go, I agree that we should document any ways we can think of where resuming the transfer can go wrong. Then we can write test cases to exercise those scenarios to prove to ourselves that we have those bases covered. I'm glad you brought up delete-after-transfer, because that is a place that could bite us if we're not being careful.

Yes, documentation and test cases will be great.  To get there, we need to work through the mental exercise to examine *every* single line of AXL code one by one, including every iteration of each for/while loop, to consider what will happen during a resume (or cancel) if the process had died just before or just after that line of code executes.  The vast majority of lines won't make a difference, but some lines will be critical, and it's hard to know which ones are important without careful consideration of every line one by one.  Easy.

src/axl.c
tonyhutter: Good catch.  I'll fix that

tonyhutter:
Latest round of changes are included in my latest push.

We should also consider adding a style checker into Travis.

src/axl.c
adammoody: Ah, I see.  In that case, I think we'd need a line here to overwrite the existing name value.
```
/* Allow caller to assign a new name */
kvtree_util_set_str(file_list, AXL_KEY_UNAME, name);
```

src/axl_io.c
```diff
@@ -494,10 +497,32 @@ int axl_meta_apply(const char* file, const kvtree* meta)
...
+ * been copied.  It's only used for test cases (like testing resuming a
+ * transfer).
+ *
+ * If the var is not set, return ULONG_MAX.
+ */
+static unsigned long axl_debug_pause_after(void)
+{
+    char* env = getenv("AXL_DEBUG_PAUSE_AFTER");
+    if (!env)
+        return ULONG_MAX;
```
adammoody, line 512:
```suggestion
    if (!env) {
        return ULONG_MAX;
    }
```

src/axl.c
```diff
@@ -280,11 +339,33 @@ int AXL_Create (axl_xfer_t xtype, const char* name)
...
+        }
+
+        if (xtype != old_xtype) {
+            AXL_ERR("Transfer type %d doesn't match transfer type %d "
+                "in state_file", xtype, old_xtype);
+            return -1;
+        }
+    } else {
+        /* We're not loading from an existing state file */
+         if (xtype == AXL_XFER_DEFAULT) {
```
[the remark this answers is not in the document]
tonyhutter: Fixed it latest push

src/axl.c
adammoody: Did you update the code for this in your push?  I think it's showing as the same for me.

adammoody: @tonyhutter , this comment came in as part of my last review, but since you didn't change it or respond in your last push, I'm guessing it got overlooked.

tonyhutter: Thanks, I did overlook that.  It's fixed in the push I just did:
```C
diff --git a/src/axl.c b/src/axl.c
index 68680ff..703e8df 100644
--- a/src/axl.c
+++ b/src/axl.c
@@ -358,7 +358,7 @@ int __AXL_Create(axl_xfer_t xtype, const char* name, const char *state_file)
             xtype = axl_detect_native_xfer();
         }
 
-        kvtree* file_list = axl_kvtrees[id];
+        file_list = axl_kvtrees[id];
         kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
         kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
         kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
```

adammoody:
@tonyhutter  or @CamStan , any idea why the travis job is stuck pending here?

tonyhutter:
I just re-pushed to kick off a new Travis test.  Hopefully that will work

CamStan:
This was happening with KVTree yesterday too, but a re-push fixed that one.

I ended up messing with some access things, that might have messed it up since some of our components are hosted on travis-ci.com and some are still on the older one, travis-ci.org.
I sent another access request that will need to be approved, so hopefully that will get it working again.

adammoody:
@tonyhutter , can you force a re-build on travis?  I just clicked a button to enable things from an email that @CamStan triggered.

adammoody:
Hmm, same outcome.  We could try a bigger hammer with an uninstall / reinstall the travis webhook.  I do see the payload URL is pointing at travis-ci.org.

CamStan:
Might as well. Travis CI's migrate to switch to travis-ci.com is only in beta right now, but we could possibly contact them to migrate if the bigger hammer doesn't work.

Also, if the bigger hammer doesn't work, could also disable the "Required status before merging" rule until we get it sorted. I believe the setting is under Settings > Branches.

CamStan:
Got them migrated and hopefully will work better now for all the components. The addresses in the READMEs, and possibly that places, will need to be updated to point to travis-ci.com now, but I can get to that eventually.

adammoody:
Thanks @tonyhutter and @CamStan !

Issue #38, opened 2019-05-22
AXL_Config


gonsie:
We should have a config function to set some options... including the number of pthead threads.

tonyhutter:
Agreed.  An `AXL_Config(id, key, value)` function could write directly into a 'config' KVtree that hangs off our main kvtree.

tonyhutter:
So I'm currently writing some "cancel a transfer" test cases, and I need a way to set config values in AXL.  Specifically, I need a 'file_delay' config value that means "in AXL_Dispatch, wait N milliseconds before copying the next file". The idea is that I can deterministically control how long my total transfer takes, so that I can call AXL_Cancel at the right time in the transfer.  I'm running into problems where my transfer is finishing before I have time to cancel it.

Here's a strawman prototype for AXL_Config:

```
/* 
 * Get/set a AXL configuration key
 *
 * This function allows you to get and set AXL configuration key values.
 *
 * key:         The configuration key you're looking up
 * set_value:   If you're setting a value, put your value here.  If you're
 *              getting a value, leave this NULL.
 *
 * On success, this will return the value you're getting or setting.  On
 * failure it will return NULL.  All values are strings.
 */
char *
AXL_Config (int id, char *key, char *set_value)
```

This will treat all configuration values as strings, which simplifies things a lot.  Yes, you'll need to convert numerical values to strings first, but I don't think that will be a big deal, as the number may already be a string anyway (like if you pass in a 'file_delay' value in argv[] to axl_cp).

Some config keys we might want:

`file_delay [milliseconds]` - Delay N milliseconds between file transfers.
`num_threads [number of threads]` - Number of threads to use for a transfer (applicable to pthreads, and whatever other xfer types we make multithreaded)
`compression [on|off]` - Tar up all the files before transfer and decompress them at the desintation.  This could be useful for node to node transfers, since scp'ing individual files can be slow.

Thoughts?

tonyhutter:
Also, should the AXL_Config values be saved to the statefile?

gonsie:
Yes, I would save these setting to the state file. 

The AXL_Config prototype looks great to me. And I think that all strings are fine. I would make the return code AXL_SUCCESS or FAILURE, not null or the value (I think you mean this, just the comment wording isn't clear).

tonyhutter:
I'd prefer it to return `NULL` or the value.  That way you can call the function directly, like:
```printf("file_delay=%s", AXL_Config(id, "file_delay", NULL));```

[...]

PR #82, opened 2020-09-01
add configuration call

rhaas80:


adammoody:
Looks good.  Thanks, @rhaas80

tonyhutter:
I feel that making the user pass a kvtree to `AXL_Config()` to set config values is needlessly complex.  It also forces the user to include KVTree in their app's dependencies.

Why not use simple config strings (`AXL_Config(id, "key1=value1 key2=value2" ...)`) as discussed in https://github.com/ECP-VeloC/AXL/issues/38?  This would allow you to set per-ID config values, and have them persist in the state_file if specified.  Also, any `AXL_Config()` code should allow the user to set the config options as many times as they want.  I bring this up because this PR only lets them set it once:

```C
    /* only accept configuration options once */
    configured = 1;
```

Also, please add a description to your PR and all your commits, and please document what the config values are and what they do.

rhaas80:
SCR does offer a string based configuration interface to allow for easy passing of user option. I saw the VeloC libraries as lower level building blocks that will be called from a higher level library that will to the option handling.

All user codes already depend (implicitly) on kvtree since AXL does.

Having that library then construct a string to pass options seems error prone since it looses any hope of type checking (not that the fact that kvtree currently stores only strings helps there).

Having said all of this: duplicating the string parsing code from SCR in each VeloC component is certainly easy enough if that is desired.

Right now the check is in place to reduce the chance of setting parameters, which AXL currently can assume to be constant during a run, to change in the middle of a run (at least to the amount that this is possible given that the config calls happen after `AXL_Init`). Can obviously be changed, and if option strings are used pretty much must be changed otherwise one has to construct one gigantic option string.

rhaas80:
Please let me know what the preferred option is and i will be happy to implement it.

rhaas80:
Yes, documentation of these options in the user docs is indeed required. Thanks for pointing that out. In the code right now the various `#define`s correspond 1:1 to existing AXL global variables that have comments (identical in .c and .h files explaining what they do). I am certainly fine triplicating these comments at he location of the `#define`.

tonyhutter:
> I saw the VeloC libraries as lower level building blocks that will be called from a higher level library that will to the option handling.

I think we should develop AXL under the assumption that it's a standalone library that anyone could use, rather than just assuming that FILO and VeloC would use it.  I could see UnifyFS using it for file stage-in, for example.

> All user codes already depend (implicitly) on kvtree since AXL does.

What I mean is that this PR would require users to `#include <kvtree.h>` and link against libkvtree if they wanted to call `AXL_Config()`.  They currently don't have to do this, nor would they have to do it with a string-based `AXL_Config()`.


> Having that library then construct a string to pass options seems error prone since it looses any hope of type checking (not that the fact that kvtree currently stores only strings helps there).

AXL would know the value type in a key=value string, so it would be able to sanity check it.  Like it would check that `file_buf_size` is passed a numeric value, for example.


> Right now the check is in place to reduce the chance of setting parameters, which AXL currently can assume to be constant during a run, to change in the middle of a run (at least to the amount that this is possible given that the config calls happen after AXL_Init). Can obviously be changed, and if option strings are used pretty much must be changed otherwise one has to construct one gigantic option string.

You wouldn't need to construct a giant string.  You could set one or more values per `AXL_Config()` call.  For example, if you wanted to set the `debug` `file_buf_size` and `mkdir` config values, you could do this:
```C
AXL_Config(id, "debug=1");
AXL_Config(id, "file_buf_size=65536 mkdir=1")
```
or this:
```C
AXL_Config(id, "debug=1");
AXL_Config(id, "file_buf_size=65536");
AXL_Config(id, "mkdir=1");
```
or this:
```C
AXL_Config(id, "debug=1 file_buf_size=65536 mkdir=1")
```
Or any other combination you want.  The result would be the same.

A string-based config is also nice because it reduces the amount of boilerplate code you have to write (which also makes the code easier to read).  For example, you could reduce your `test_config.c` code from:

```C
#include <kvtree.h>
...

     kvtree* axl_config_values = kvtree_new(); 
...

    /* check AXL configuration settings */
    rc = kvtree_util_set_bytecount(axl_config_values,
                                   AXL_KEY_CONFIG_FILE_BUF_SIZE,
                                   old_axl_file_buf_size + 1);
    if (rc != KVTREE_SUCCESS) {
        printf("kvtree_util_set_bytecount failed (error %d)\n", rc);
        return rc;
    }
    rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_DEBUG,
                             !old_axl_debug);
    if (rc != KVTREE_SUCCESS) {
        printf("kvtree_util_set_int failed (error %d)\n", rc);
        return rc;
    }
    rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_MKDIR,
                             !old_axl_make_directories);
    if (rc != KVTREE_SUCCESS) {
        printf("kvtree_util_set_int failed (error %d)\n", rc);
        return rc;
    }
    rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_COPY_METADATA,
                             !old_axl_copy_metadata);
    if (rc != KVTREE_SUCCESS) {
        printf("kvtree_util_set_int failed (error %d)\n", rc);
        return rc;
    }

    rc = AXL_Config(axl_config_values);

    ...
    kvtree_delete(&axl_config_values);
```


To:

```C
	char *config;
	asprintf(&config, "file_buf_size=%d debug=%d mkdir=%d copy_metadata=%d",
		old_axl_file_buf_size + 1, !old_axl_debug, !old_axl_make_directories, !old_axl_copy_metadata);
	AXL_Config(config);
```

I would also propose that `AXL_Config()` be able to set and retrieve both global and per AXL ID specific values.  There are a couple of ways to do this:

1. Pass an AXL ID value to `AXL_Config()`.  If it's a global value, then pass -1 for the AXL ID:

```C
int id = 42;
AXL_Config(id, "file_buf_size=65535");
AXL_Config(-1, "debug=1");
````

2. Have an 'id' key value pair for setting ID specific values:

```C
AXL_Config("id=42 file_buf_size=65535");
AXL_Config("debug=1");
```

And to retrieve:
```C
char *str = AXL_Config("debug");
printf("debug=%s", str);   // prints "debug=1"

char *str = AXL_Config("id=42 file_buf_size");
printf("file_buf_size=%s", str);   // prints "file_buf_size=65535"
```

rhaas80:
Uff that was a lengthy comment. I guess I will have to redo my implementation.

> I think we should develop AXL under the assumption that it's a standalone library that anyone could use, rather than just assuming that FILO and VeloC would use it. I could see UnifyFS using it for file stage-in, for example.

From my point of view (as a science code developer) UnifyFS would still be a layer between AXL and "me" and would (and should do so I would think) hide the fact that it internally uses AXL and translate its own options into whatever AXL expects.

> What I mean is that this PR would require users to `#include <kvtree.h>` and link against libkvtree if they wanted to call `AXL_Config()`. They currently don't have to do this, nor would they have to do it with a string-based `AXL_Config()`.

Right now one already has to *link* against kvtree since AXL links against it. One has to explicitly do so since the current CMake file for AXL is not reporting AXL's own dependent libraries as libraries one needs to link against (at least not that I can see and in a similar situation in er with rankstr I certainly had to manually add it). On a sufficiently old system with dynamic linking the linker will automatically link in the dynamic librraries that libAXL.so uses when one links against libAXL.so but for static linking and newer linkers (that default to not setting ``--copy-dt-needed-entries`` see https://stackoverflow.com/questions/8725572/gcc-4-5-vs-4-4-linking-with-dependencies) one does need to list them even for a dynamic link.

`#include <kvtree.h>` is not needed on the user side since `axl.h` already includes that file. *However* this requires that include paths are set up so that the code using AXL actually finds `kvtree.h` which is a downside and could be avoided by forward declaring `kvtree` ie `typedef struct kvtree_struct      kvtree;`.

> > Having that library then construct a string to pass options seems error prone since it looses any hope of type checking (not that the fact that kvtree currently stores only strings helps there).
> 
> AXL would know the value type in a key=value string, so it would be able to sanity check it. Like it would check that `file_buf_size` is passed a numeric value, for example.

Let me be more explicit: passing strings pushes type checks to runtime, while using `kvtree_set_int` gives a warning at runtime. One can get warnings for strings by using something like `AXL_Configf` that has printf like syntax and using compiler specific attributes to tell the compiler that `ACL_Confgf` is printf-like.

> You wouldn't need to construct a giant string. You could set one or more values per `AXL_Config()` call. For example, if you wanted to set the `debug` `file_buf_size` and `mkdir` config values, you could do this:

If I look at the discussion in #38 (eg https://github.com/ECP-VeloC/AXL/issues/38#issuecomment-502798719) then this does not seem to be (sanely) possible since constructs like "CKPT=0 SCHEME=XOR" are not expected to set both `CKPT` to 0 and `SCHEME` to `XOR` but instead set `SCHEME` to `XOR` in the `0` sub-option of `CHKPT`. Admittedly none of the VeloC compoments actually uses anything like this (only SCR does).

> A string-based config is also nice because it reduces the amount of boilerplate code you have to write (which also makes the code easier to read). For example, you could reduce your `test_config.c` code from:
> 
> ```c
> #include <kvtree.h>
> ...
> 
>      kvtree* axl_config_values = kvtree_new(); 
> ...
> 
>     /* check AXL configuration settings */
>     rc = kvtree_util_set_bytecount(axl_config_values,
>                                    AXL_KEY_CONFIG_FILE_BUF_SIZE,
>                                    old_axl_file_buf_size + 1);
>     if (rc != KVTREE_SUCCESS) {
>         printf("kvtree_util_set_bytecount failed (error %d)\n", rc);
>         return rc;
>     }
>     rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_DEBUG,
>                              !old_axl_debug);
>     if (rc != KVTREE_SUCCESS) {
>         printf("kvtree_util_set_int failed (error %d)\n", rc);
>         return rc;
>     }
>     rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_MKDIR,
>                              !old_axl_make_directories);
>     if (rc != KVTREE_SUCCESS) {
>         printf("kvtree_util_set_int failed (error %d)\n", rc);
>         return rc;
>     }
>     rc = kvtree_util_set_int(axl_config_values, AXL_KEY_CONFIG_COPY_METADATA,
>                              !old_axl_copy_metadata);
>     if (rc != KVTREE_SUCCESS) {
>         printf("kvtree_util_set_int failed (error %d)\n", rc);
>         return rc;
>     }
> 
>     rc = AXL_Config(axl_config_values);
> 
>     ...
>     kvtree_delete(&axl_config_values);
> ```
> 
> To:
> 
> ```c
> 	char *config;
> 	asprintf(&config, "file_buf_size=%d debug=%d mkdir=%d copy_metadata=%d",
> 		old_axl_file_buf_size + 1, !old_axl_debug, !old_axl_make_directories, !old_axl_copy_metadata);
> 	AXL_Config(config);
I don't know. This is lacking a `free` on `config`. It does not check asprintf for failure. This seems to mostly safe on boilerplate by removing error checks. With full error checks with strings the code and trying to avoid typo errors in the string names of the variables would look like:

```
char *config;

rc = asprintf(&config, AXL_CONFIG_KEY_FILE_BUF_SIZE"=%lu "AXL_CONFIG_KEY_DEBUG"=%d "AXL_CONFIG_KEY_MKDIR"=%d "AXL_CONFIG_KEY_COPY_METADATA"=%d",
  (unsigned long) old_axl_file_buf_size + 1, !old_axl_debug, !old_axl_make_directories, !old_axl_copy_metadata);
if(rc == -1) {
  printf("asptrinf failed (error %d)\n", rc);
  return 1;
}
rc = AXL_Config(config);
if(rc != AXL_SUCCESS) {
  return 1;
}
free(config);
```

which is still quite a bit shorter since it combines multiple `kvtree_set_XXX` calls into a single `asprintf`.

> I would also propose that `AXL_Config()` be able to set and retrieve both global and per AXL ID specific values. There are a couple of ways to do this:

sure.

I will implement a demo setup and let you have a look.

tonyhutter:
Thanks for the info.  Since we decided to go with the kvtree-approach in today's meeting, you can ignore my previous comments, and we should probably update https://github.com/ECP-VeloC/AXL/issues/38 to say we're using a KVTree.

As far as per-AXL ID vs global values go, may I suggest we mix both globals and ID specific values in one tree, with the ID specific ones in a subtree, like:
```C
// psudocode: set global debug and some ID specific values
kvtree config = {
	debug = 1
	kvtree {
		id = 42
		file_buf_size = 65536
	}
	kvtree {
		id = 43
		file_buf_size = 2097152
		compression = 1
	}
}

AXL_Config(config)
```
I think there's even a way you could index the subtree by `id` if you wanted, but I can't remember off the top of my head.

And for reading config values back, its probably easiest to have a way to return all the config values (global and ID-specific), and not worry about querying individual ones.  You could extend `AXL_Config()` to do that:
```C
/*
 * Get/set AXL configuration values.
 *
 * config:	The new configuration.  Global variables are in top level of
 *              the tree, and per-ID values are subtrees.  If config=NULL,
 *              then return a kvtree with all the configuration values (globals
 *              and all per-ID trees).
 *
 * Return value:	If config != NULL, then return config on success.  If
 *                      config=NULL (you're querying the config) then return
 *                      a new kvtree on success.  Return NULL on any failures.
 */
kvtree * AXL_Config(kvtree *config)
```
For example:
```C
kvtree *old_config;

old_config = AXL_Config(NULL);
if (!old_config)
	printf("error\n");
// read back our config values into old_config
```

In fact, you could have return not only the config values, but all kvtree values for a particular ID too (like AXL_KEY_STATUS, AXL_KEY_STATE, ... etc).

rhaas80:
Keeping track of things, @tonyhutter pointed out https://github.com/ECP-VeloC/AXL/pull/82#issuecomment-689742336 with the choice of what to return for setting (the input config) and getting (a full tree of all settings).

src/axl.c
```diff
@@ -309,6 +310,124 @@ int AXL_Finalize (void)
...
+            kvtree_elem_key(elem),
+            kvtree_elem_key(kvtree_elem_first(kvtree_elem_hash(elem)))
+          );
+          retval = NULL;
+        }
+      }
+    }
+
+    /* only accept configuration options once */
+    configured = 1;
```
tonyhutter, line 421: Regarding the "only accept configuration options once", you had earlier mentioned:

> Right now the check is in place to reduce the chance of setting parameters, which AXL currently can assume to be constant during a run, to change in the middle of a run (at least to the amount that this is possible given that the config calls happen after AXL_Init()

I don't think we can make that assumption, since we can't predict what the user will want to do, nor what config options we may want to add in the future.  It also prevents the user from setting different config options at different spots in their code.   For example, a user couldn't set AXL_KEY_CONFIG_DEBUG in their "init" code and then later set  AXL_KEY_CONFIG_MKDIR in their "parsing my command line options" code.  My vote would be to let the user call AXL_Config() as many times as they want.

test/test_config.c
```diff
@@ -0,0 +1,99 @@
...
+    int rc;
+    char *state_file = NULL;
+    kvtree* axl_config_values = kvtree_new();
+
+    size_t old_axl_file_buf_size = axl_file_buf_size;
+    int old_axl_debug = axl_debug;
+    int old_axl_make_directories = axl_make_directories;
+    int old_axl_copy_metadata = axl_copy_metadata;
+
+    rc = AXL_Init(state_file);
```
tonyhutter, line 20:
```suggestion
    rc = AXL_Init(NULL);
```
...and remove `state_file`.  `AXL_Init(state_file)` is deprecated.

test/test_config.c
```diff
@@ -0,0 +1,99 @@
...
+    if (rc != KVTREE_SUCCESS) {
+        printf("kvtree_util_set_int failed (error %d)\n", rc);
+        return rc;
+    }
+
+    printf("Configuring AXL...\n");
+    if (AXL_Config(axl_config_values) == NULL) {
+        printf("AXL_Config() failed\n");
+        return EXIT_FAILURE;
+    }
```
tonyhutter, line 57: You need to AXL_Create() a transfer and add tests for per-ID config options.

src/axl.h
```diff
@@ -24,6 +24,15 @@ extern "C" {
  *  \ingroup axl
  *  \brief asynchronous transfer library */
 
+/** AXL configuration options */
+#define AXL_KEY_CONFIG_FLUSH_ASYNC_BW "FLUSH_ASYNC_BW" /* UNUSED */
+#define AXL_KEY_CONFIG_FLUSH_ASYNC_PERCENT "FLUSH_ASYNC_PERCENT" /* UNUSED */
+#define AXL_KEY_CONFIG_FILE_BUF_SIZE "FILE_BUF_SIZE"
+#define AXL_KEY_CONFIG_CRC_ON_FLUSH "CRC_ON_FLUSH" /* UNUSED */
```
tonyhutter, line 31: Remove all `/* UNUSED */` values.

src/axl.c
```diff
@@ -309,6 +310,124 @@ int AXL_Finalize (void)
...
+        }
+      }
+
+      kvtree_util_get_double(config,
+        AXL_KEY_CONFIG_FLUSH_ASYNC_PERCENT, &axl_flush_async_percent);
+
+      if (kvtree_util_get_bytecount(config,
+        AXL_KEY_CONFIG_FILE_BUF_SIZE, &ul) == KVTREE_SUCCESS)
+      {
+        axl_file_buf_size = (int) ul;
```
tonyhutter, line 366: `axl_file_buf_size` is a `size_t` (which most likely is an `unsigned long` on 64-bit), so you can remove the check against the `int`.  Or if you're being super-safe, check `ul` against `SIZE_MAX`.

tonyhutter:
Please check that all indents are four spaces, and also add per-ID config options.

tonyhutter:
Please also squash your commits

test/test_config.c
rhaas80: At this point there are no per-ID config options so nothing to test.

test/test_config.c
rhaas80: Done.

src/axl.c
rhaas80: I changed the code as suggested.

src/axl.c
rhaas80: sorry, I had left the fix in rhaas/keyvalue_get. Cherry picked into rhaas/keyvalue. I'd rather not make assumptions about what `size_t` is wrt to `unsigned long`. The smallest type to be guaranteed by C to be 64bit is `long long int` with `long` being allowed to be 32bit.

src/axl.h
rhaas80: Removed as requested.

src/axl.c
tonyhutter: Then I would just make `axl_file_buf_size` an `unsigned long` and remove the check.

src/axl.c
```diff
@@ -309,6 +310,92 @@ int AXL_Finalize (void)
     return rc;
 }
 
+
+/** Set a AXL config parameters */
+kvtree* AXL_Config(const int id, const kvtree* config)
```
tonyhutter, line 315: I would get rid of the `id` arg, and make the per-ID values subtrees within the main tree.  For example:

```C
    kvtree *tree;
    kvtree *id;

    tree = kvtree_new();

    /* Add some config for AXL ID 1 */
    id = kvtree_set_kv_int(tree, "id", 1);
    kvtree_set_kv_int(id, "compression", 1);
    kvtree_set_kv_int(id, "file_buf_size", 65536);

    /* Add some config for AXL ID 42 */
    id = kvtree_set_kv_int(tree, "id", 42);
    kvtree_set_kv_int(id, "file_buf_size", 1048576);

    printf("Whole tree ---\n");
    kvtree_print(tree, 4);

    /* Lookup an id 1's values */
    id = kvtree_get_kv_int(tree, "id", 1);
    printf("Just ID 1 ---\n");
    kvtree_print(id, 4);
```
```
Whole tree ---
      id
        42
          file_buf_size
            1048576
        1
          file_buf_size
            65536
          compression
            1
Just ID 1 ---
      file_buf_size
        65536
      compression
        1
```

test/test_config.c
tonyhutter: `axl_file_buf_size` should be per-ID

src/axl.c
rhaas80: This (changing option at runtime in particular after having created a transfer) is affected by #85 since AXL also accesses the global values of the configuration options without protecting them between multiple threads. Right now the best solution seems to me to make copies of the global variables when a new transfer is created and initialize the transfer-local copies with the global values. This way there is consistent behaviour when changing configuration options at runtime namely a newly created AXL transfer will use the global options active at the time it was created (rather than whatever value the global options had at the time the thread driving the transfer reads them out).

tonyhutter: > Right now the best solution seems to me to make copies of the global variables when a new transfer is created and initialize the transfer-local copies with the global values. 

Then they don't behave as globals anymore.   A better solution is to protect the globals with a reader-writer lock.

rhaas80: There seems to be a misunderstanding. I am proposing to initialize the per-transfer options with the values of the global options at the time the transfer is created. AXL in master will always use the current value of the global variables. This pull request makes these global variables that hold the configuration values non-const in between `AXL_Init` and the end of the code which means the observed behaviour is that the behaviour of a transfer is a race condition between when the transfer thread looks at the global variables and when the user code calls `AXL_Config`. 
A possible way around, that I had not considered before, is to forbid calls to `AXL_Config` after the first transfer has been started thought that does not seem to be what was requested.

The global variables will need to be protected by a lock if they are used in the threads, which is, at the very least potentially true for the `axl_debug` option. This would be in addition to making a copy of their values when creating the transfer. Simiarly the function that modify per-transfer options need to get a lock before they modify the transfer options and the threads using them need to use the same lock.

I am admitting to confusion why a global option not being constant would make it not global anymore.

tonyhutter: >  I am proposing to initialize the per-transfer options with the values of the global options at the time the transfer is created. 

Sorry, I'm confused.  Are you saying the defaults for the per-transfer config values would be configurable as globals?  For example, having a configurable global `file_buf_size` which the per-transfer `file_buf_size` values get initially set to?

rhaas80: Well mostly. I am not really saying "would be". I am saying "this is what AXL does in its master branch". Namely there is a global variable `axl_file_buf_size` which is initialized to `(size_t) 1048576` in `__AXL_Init` (and written to only there, it is not changed anywhere else). 

This global variable is then used later on eg as an argument to `axl_file_copy`: `axl_file_copy(source, destination, axl_file_buf_size, resume)` by all transfers.

The configuration interface will change this global variable if the "global" ID is passed to it.

Thus when allowing per transfer configurable `file_buf_size` it seem natural that if a user does not explicitly configure `file_buf_size` then the value of `axl_file_buf_size` is used since this is the behaviour in `master` (where no configuration is possible). 

Does this help clarify what I tried to say?

adammoody: Looks good to me.  Until we modify the code to change config settings on a per transfer basis, this lets the AXL user to change hard coded values.  Per transfer settings is another level up and can be added in a later PR.

tonyhutter: @adammoody that's fine, we can tack on per-transfer config options later.

src/axl.c
rhaas80: Uh, I have a not quite pull request (but a branch) that actually implements per transfer options. No thread locking yet. The diff can be seen here:

https://github.com/rhaas80/AXL/compare/rhaas/keyvalue_get...rhaas80:rhaas/per_transfer

and the branch itself is

https://github.com/rhaas80/AXL/tree/rhaas/per_transfer

and the second last commit (https://github.com/rhaas80/AXL/commit/b6a170077a576fa8cfca1c42eba4623449848d1c) is the one that goes from `id` being an argument to encoding it in the kvtree. I am not sure I *like* that solution, but it is certainly doable and the commit nicely shows the differences this implies to taking an `int id` argument in `AXL_Config`.

rhaas80: in call on 2020-10-07:
* go with `kvtree` only argument test
* combine all 3 pull requests for `set`, `get` and `per_transfer` into single pull request
* might keep commits separate

src/axl.h
```diff
@@ -82,6 +88,23 @@ int AXL_Finalize (void);
...
+ * config: The new configuration.  Global variables are in top level of
+ *              the tree, and per-ID values are subtrees.  If config=NULL,
+ *              then return a kvtree with all the configuration values (globals
+ *              and all per-ID trees).
+ *
+ * Return value: If config != NULL, then return config on success.  If
+ *                      config=NULL (you're querying the config) then return
+ *                      a new kvtree on success.  Return NULL on any failures.
+ */
+typedef struct kvtree_struct kvtree;
```
tonyhutter, line 103: `kvtree` is already `typedef`'d in kvtree.h.

src/axl.c
```diff
@@ -309,6 +310,248 @@ int AXL_Finalize (void)
...
+{
+    kvtree* config = kvtree_new();
+    assert(config != NULL);
+
+    static const char* known_options[] = {
+      AXL_KEY_CONFIG_FILE_BUF_SIZE,
+      AXL_KEY_CONFIG_MKDIR,
+      AXL_KEY_CONFIG_COPY_METADATA,
+      NULL
+    };
```
tonyhutter, line 485: Indents should be 4 spaces

tonyhutter:
Double check that your indents are all four spaces.

src/axl.h
rhaas80: The point here, as requested earlier in this ticket, is to not require user to include `kvtree.h` is they are not making use of `AXL_Config` hence I must forward declare the type. DId the `#include "kvtree.h"` line creep back in.

src/axl.c
rhaas80: I was reluctant to change types of existing variables given that I do now know here they are used. If you think this is safe, fine with me for sure (makes for easier code). Note that right now `kvtree_util_get_bytecount` and `scr_abtoull` which SCR (the original code) uses to convert byte count strings to integer variables are inconsistent. The former using `unsigned long` (which is only guaranteed to be 32bit and is indeed so even on 64bit windows machines) while the latter uses `unsigned long long` (which is the smallest type guaranteed to be able to hold a 64bit number). So unless those two are changed there will still be a need of checks (though it could / should actually go into `kvtree_util_get_bytecount` making `kvtree`'s user code much simpler).

rhaas80:
> Double check that your indents are all four spaces.
Will do.

I believe i asked this before: is there an actual document (or even better, a clang format file and an offical clang version to use) for the desired coding style? Right now SCR uses 2 spaces, AXL spaces 4 and there may well be subtle other differences.

src/axl.h
tonyhutter: Ah you're right, I assumed you had `#include <kvtree.h>` in `axl.h` for the `struct kvtree_struct` definition, but the way you typedefed it gets around needing that definition unless you use it.  Very cool trick! :+1:

tonyhutter:
Yea, the inconsistency is annoying, and no, there's no style docs.  We've talked a little about enforcing style in Travis: 
https://github.com/ECP-VeloC/AXL/pull/77#issuecomment-678553793 but haven't done it.

tonyhutter (approved):
Looks good!

rhaas80:
Based on the discussion in today's call I verified that with AXL at rhaas/keyvalue (ie this pull request) and all other VeloC components on master and SCR on develop, no more tests fail on my workstation than with AXL on master (some tests fail due to they assuming SLURM and / or multiple nodes).

doc/README.md
```diff
@@ -19,6 +19,49 @@ If no state files is needed, one may pass NULL in place of the path name.
...
+
+axl.h defines symbolic names for all parameter which should be used instead of
+the immediate strings whenever possible to avoid typos and runtime errors.
+
+The current set of configuariont options including their default value is
+(prefix Name by AXL\_KEY\_CONFIG\_ for symbolic name):
+
+Name | Type  | Default | Per transfer | Description
+-----|------:|-----:|:-:|--------------------------------------------------------
+FILE\_BUF\_SIZE | Byte count | 1048576 | Yes | Specify the number of bytes to use for internal buffers when copying files between the parallel file system and the cache.
```
adammoody, line 60:
```suggestion
FILE\_BUF\_SIZE | Byte count | 1048576 | Yes | Specify the number of bytes to use for internal buffers when copying files between source and destination.
```

doc/README.md
```diff
@@ -19,6 +19,49 @@ If no state files is needed, one may pass NULL in place of the path name.
...
+DEBUG           |    Boolean |       0 |  No | Set to 1 to have AXL print debug messages to stdout, set to 0 for no output.
+MKDIR           |    Boolean |       1 | Yes | Specifies whether the device supports the creation of directories (1) or not (0).
```
adammoody, line 62:
```suggestion
MKDIR           |    Boolean |       1 | Yes | Specifies whether the destination file system supports the creation of directories (1) or not (0).
```

PR #84, opened 2020-09-30
implement config get interface

rhaas80:
This pull request implements code to actually return values on `AXL_Config(NULL)`. Please take a look if this is sane and I will implement the same logic in the other components.

Things to note:

* there are some parameters that are accepted by `AXL_Config` to set but are ignored (documented but not used), these will not show up in the return `kvtree`

* the returned kvtree must be freed

* `test_config` now also checks the query interface

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -362,7 +362,7 @@ static kvtree* AXL_Config_Set(const kvtree* config)
       if (kvtree_util_get_bytecount(config,
         AXL_KEY_CONFIG_FILE_BUF_SIZE, &ul) == KVTREE_SUCCESS)
       {
-        axl_file_buf_size = (int) ul;
+        axl_file_buf_size = (size_t) ul;
         if (axl_file_buf_size != ul) {
           char* value;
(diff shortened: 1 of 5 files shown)


tonyhutter:
In github, you can mark this PR as a based off of your `rhaas80:rhaas/keyvalue` branch.  That will make it so we don't have to see all the old AXL_Config() commits combined with the commits from this PR.  Or just combine this branch into your other PR, and have one big AXL_Config() PR.

rhaas80:
I do not seem to be able to base this off my `rhaas/keyvalue` branch since at least as far GItHub's "merge into" option goes. Maybe I am missing something.  My guess would be that this because `rhaas/keyvalue` is in my own fork of the repository but the pull request always appears in the target repository (ie LLNL's) where there is no `rhaas/keyvalue` branch. It might be simpler to first get the setter part into LLNL's repository then I can rebase the getter against it and the diff in the pull request will become more mananable.

rhaas80:
combined with "set" pull request https://github.com/ECP-VeloC/AXL/pull/84

PR #86, opened 2020-11-10
Add AXL_XFER_STATE_FILE xfer type

tonyhutter:
Add `AXL_XFER_STATE_FILE` transfer type.  This type is used to tell AXL that you want to use the transfer type specified in the
state_file.  This is useful if you're trying to resume a transfer from a previous state_file, since it saves you from having to
remember the old transfer type you used.

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -577,9 +577,14 @@ int __AXL_Create(axl_xfer_t xtype, const char* name, const char* state_file)
         kvtree* file_list = NULL;
         axl_xfer_t old_xtype = AXL_XFER_NULL;
         axl_xfer_state_t old_xstate = AXL_XFER_STATE_NULL;
+
         if (axl_get_info(id, &file_list, &old_xtype, &old_xstate) != AXL_SUCCESS) {
             AXL_ERR("Couldn't get kvtree info");
             return -1;
+        }
+
+        if (xtype == AXL_XFER_STATE_FILE) {
+            xtype = old_xtype;
         }
 
         if (xtype != old_xtype) {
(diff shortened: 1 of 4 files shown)


src/axl.c
```diff
@@ -346,34 +346,38 @@ int __AXL_Create(axl_xfer_t xtype, const char* name, const char* state_file)
...
+        xtype = axl_detect_default_xfer();
+    } else if (xtype == AXL_XFER_NATIVE) {
+        xtype = axl_detect_native_xfer();
+    } else if (xtype == AXL_XFER_STATE_FILE && !reload_from_state_file) {
+        AXL_ERR("Can't use AXL_XFER_STATE_FILE without a state_file");
+        return -1;
     }
 
+    kvtree* file_list = axl_kvtrees[id];
+    kvtree_util_set_int(file_list, AXL_KEY_XFER_TYPE, xtype);
```
adammoody, line 376: If ``xtype == AXL_XFER_STATE_FILE`` at this point, won't it save that as the transfer type in the file, so that if one resumes again we'd get that special value instead of the actual transfer type?

adammoody: Nope, nevermind.  I see why that's not the case now.

src/axl.c
```diff
@@ -346,34 +346,38 @@ int __AXL_Create(axl_xfer_t xtype, const char* name, const char* state_file)
...
+    kvtree_util_set_str(file_list, AXL_KEY_UNAME, name);
+    kvtree_util_set_int(file_list, AXL_KEY_STATUS, AXL_STATUS_SOURCE);
+    kvtree_util_set_int(file_list, AXL_KEY_STATE, (int)AXL_XFER_STATE_CREATED);
```
adammoody, line 379: This block of code used to only execute when not reloading from a state file, and now it always executes.  I haven't through this all the way through.  Is it always safe to reset these fields?  If the transfer has already started, I think it could have some other state, like DISPATCHED or COMPLETED.

tonyhutter: Good catch, I think you're right.  I believe `AXL_STATUS_SOURCE` and `AXL_KEY_STATE` should be kept as-is in the state_file case.

tonyhutter: Fixed in latest push

tonyhutter:
I just rebased on master, I believe this is good-to-go.

PR #87, opened 2020-11-17
AXL: protect AXL's transfer file lists in multi-threading transfers

rhaas80:
This is a continuation / transfer of https://github.com/rhaas80/AXL/pull/1 please see there for some comments.

Implementation notes:

* this assumes that AXL itself is not being called from a multi-threaded code, something already (tacitly) assumed by the way the state file is handled but not explicitly spelled out
* calling `AXL_Config` on a transfer that has already been dispatch is a race condition (since there is no lock on the individual `file_list` kvtrees) and not supported
* calling `AXL_Config` for the DEBUG settings is race condition (since `axl_debug` is not protected but is accessed from within the worker threads) while there are already dispatched transfers and will not be fixed

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -62,6 +62,11 @@ DEBUG           |    Boolean |       0 |  No | Set to 1 to have AXL print debug 
 MKDIR           |    Boolean |       1 | Yes | Specifies whether the destination file system supports the creation of directories (1) or not (0).
 COPY\_METADATA  |    Boolean |       0 | Yes | Whether file metadata like timestamp and permission bits should also be copied.
 
+Thread safety: setting the DEBUG or any per-transfer configuration value after
+the transfer has been dispatched entails a race contion between the main thread
+and the worker threads. Changing configuration options after a transfer has
+been dispatched is not supported.
+
 # Transferring files
 Regardless of the transfer type, the basic control flow of a transfer is always:
 1. AXL\_Create - allocate a new transfer object, providing its type and a name
--- a/src/axl.c
+++ b/src/axl.c
@@ -3,7 +3,6 @@
 #include <string.h>
 #include <errno.h>
 #include <stdio.h>
-#include <pthread.h>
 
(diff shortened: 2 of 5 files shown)


tonyhutter:
Sorry I've taken so long to look at this, I was out the past two weeks.  So basically we're going from "AXL is somewhat, but not completely, thread safe" to telling the user "AXL is not thread safe" and removing the locking.  I think that's fine, at least we're consistent now.

tonyhutter (approved):
Looks good, please squash and rebase your commits.

rhaas80:
Thank you for the review.

I have rebased onto current master, and squashed all three commits.

adammoody:
@rhaas80 , while merging other PRs, I found this outstanding PR you had opened.  I haven't had time to look at it in detail.

Is this still good to go?

Does this address the realloc race condition you had spotted?

rhaas80:
Well for *some* definition of addresses the issue. Tony summarizes the change correctly in https://github.com/ECP-VeloC/AXL/pull/87#issuecomment-736803323 and basically we are going from "we claim to everywhere be but aren't really" to "we explicitly say we aren't in the user facing function calls but otherwise we are internally" thread safe.

So it removes the bug since all access by the transfer thread is now thread safe (it keeps its own copy of the data item) while we explicitly state that calls to `AXL_XXX` functions from the client side must not be thread-concurrent ie the client must serialize them (though can call the functions from multiple threads).

adammoody:
Awesome.  Thanks, @rhaas80 .  We should work to get this merged in that case.  I see we've got some merge conflicts on it now.  Would you mind refreshing it?

rhaas80:
Done the conflict was due to 7fa8f9e which introduced
```C
static int bbapi_is_loaded = 0;
```
too close to where i removed a lock variable so get had problems merging.

adammoody:
Thanks @rhaas80 !

PR #90, opened 2021-01-08
drop -Werror from default build flags

adammoody:
One of our users tripped on a ``-Werror`` message.

```
axl_sync.c:39:13 error: variable 'success' set but not used [-Werror=unused-bug-set-variable]
  int success;
```

This PR disables those to make life easier for end users building with newer compilers that may be more strict.

We should use these flags in our travis/testing/development builds, but let's otherwise make life easier for our end users.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,8 +18,7 @@ OPTION(AXL_LINK_STATIC "Default to static linking? (Needed for Cray)" OFF)
 SET(AXL_ASYNC_API "NONE" CACHE STRING "Vendor-specific asynchronous file transfer (CRAY_DW INTEL_CPPR IBM_BBAPI NONE)")
 SET_PROPERTY(CACHE AXL_ASYNC_API PROPERTY STRINGS NONE CRAY_DW INTEL_CPPR IBM_BBAPI)
 
-
-SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Werror")
+SET(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall")
 
 # Find Packages & Files
 


CamStan (approved):
Any idea what compiler was used? We could create an issue to expand the compilers used by Travis/Gitlab if we want to catch these warnings/errors.

tonyhutter:
AXL builds normally with gcc 10.2, which is the newest stable.   The user is probably building with `cmake -DCMAKE_BUILD_TYPE=Release ..` which disables the `asserts()`.   It's getting hung up on variables that are declared and only used for asserts:
```
Scanning dependencies of target axl_o
[  7%] Building C object src/CMakeFiles/axl_o.dir/axl.c.o
/home/hutter2/AXL/src/axl.c: In function ‘AXL_Config_Set’:
/home/hutter2/AXL/src/axl.c:399:31: error: unused variable ‘kvtree_first_elem_hash’ [-Werror=unused-variable]
  399 |                 const kvtree* kvtree_first_elem_hash =
      |                               ^~~~~~~~~~~~~~~~~~~~~~
/home/hutter2/AXL/src/axl.c:438:23: error: unused variable ‘kvtree_first_elem_hash’ [-Werror=unused-variable]
  438 |         const kvtree* kvtree_first_elem_hash =
      |                       ^~~~~~~~~~~~~~~~~~~~~~
/home/hutter2/AXL/src/axl.c: In function ‘__AXL_Dispatch’:
/home/hutter2/AXL/src/axl.c:953:9: error: unused variable ‘success’ [-Werror=unused-variable]
  953 |     int success = kvtree_util_get_int(file_list, AXL_KEY_CONFIG_MKDIR,
      |         ^~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [src/CMakeFiles/axl_o.dir/build.make:82: src/CMakeFiles/axl_o.dir/axl.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:979: src/CMakeFiles/axl_o.dir/all] Error 2
make: *** [Makefile:160: all] Error 2
```
My vote would be to fix these build issues and leave `-Wall` on, but I'll bow to @adammoody for the final say.

PR #91, opened 2021-01-08
drop macro mangling for axl interface

adammoody:
Though this trick allows users of the old interface to compile, it still leaves them broken since it does not function according to the old semantics.  Let's bite the bullet and make users upgrade to the new interface.

### DIFF
--- a/src/axl.h
+++ b/src/axl.h
@@ -52,19 +52,10 @@ typedef enum {
     AXL_XFER_STATE_FILE,    /* Use the xfer type specified in the state_file. */
 } axl_xfer_t;
 
-#define ARG0(dummy, a0, ...) a0
-#define GET_ARG0(...) ARG0(dummy, ## __VA_ARGS__, 0)
-
 /*
  * int AXL_Init(void) - Initialize the library.
- *
- * NOTE: AXL_Init() used to take in a state_file argument, but this has
- * since been removed. We do some macro mangling to make sure both state_file
- * version and the no-arg version both work, but the state_file version is
- * deprecated and should not be used.
  */
-#define AXL_Init(...) __AXL_Init(GET_ARG0(__VA_ARGS__))
-int __AXL_Init (const char* state_file);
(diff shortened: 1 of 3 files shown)


src/axl.h
```diff
@@ -78,16 +72,8 @@ int AXL_Finalize (void);
...
- * NOTE: AXL_Create() used to only take in type and name.  The state_file arg
- * was added later.  We do some macro mangling to make sure both 2-arg and
- * 3-arg versions of AXL_Create() work, but the 2-arg version is deprecated
- * and should not be used.  If you're loading from an existing state_file, then
- * type must be the same as the type in state_file.
  */
-#define AXL_Create(type, name, ...) \
-        __AXL_Create(type, name, GET_ARG0(__VA_ARGS__))
-int __AXL_Create (axl_xfer_t xtype, const char* name, const char* state_file);
+int AXL_Create (axl_xfer_t xtype, const char* name, const char* state_file);
```
CamStan, line 76: Does this mean anyone still using the 2-arg version of `AXL_Create()`, they will now have issues?

adammoody: Yes, this change will break anyone with an ``AXL_Create(type, name)``, so they'll need to update to ``AXL_Create(type, name, NULL)``.  We'll need to update veloc and wemul.

CamStan: That's what I wanted to check. Thought @fahimcsebuet might want to know this change was happening. 😄

CamStan (approved):

tonyhutter (approved):

src/axl.h
adammoody: @fahimcsebuet , as @CamStan noted, whenever you update AXL, you'll need to tweak your AXL_Init and AXL_Create lines from:
```
AXL_Init(NULL);
AXL_Create(type, name);
```
to
```
AXL_Init();
AXL_Create(type, name, NULL);
```

Issue #70, opened 2020-06-10
Don't call axl_async_init_bbapi() in AXL_Init()


tonyhutter:
If you build AXL on a machine that has the BBAPI headers/libraries, but is not running the BB software itself, you will get this error when doing any type of transfer:
```
$ ./test/axl_cp /tmp/file1 /tmp/file2
AXL 0.3.0 ERROR: butte5: AXL Error with BBAPI rc:       -2 @ bb_check /g/g0/hutter2/AXL/src/axl_async_bbapi.c:131
AXL 0.3.0 ERROR: butte5: AXL Error with BBAPI details:  {"id":"1","rc":"-2","env":{"configfile":"\/etc\/ibm\/bb.cfg"},"error":{"text":"Unable to create bb.proxy connection","func":"BB_InitLibrary","line":"238","sourcefile":"\/u\/tgooding\/cast_proto1.5\/CAST\/bb\/src\/bbapi.cc"}} @ bb_check /g/g0/hutter2/AXL/src/axl_async_bbapi.c:132
AXL_Init() failed (error -1)
```
This happens because we're calling `axl_async_init_bbapi()` in `AXL_Init()`.  We should defer the `axl_async_init_bbapi()` until `AXL_Create()`.

tonyhutter:
This is affecting my work on the `scr_poststage` script.  Specifically, I'm getting errors when I'm finalizing a previously completed transfer from the IBM login node (which has the BB API headers, but is not running the client).  When I call `AXL_Finalize()`,  I get:
```
AXL 0.3.0 ERROR: lassen708: AXL Error with BBAPI rc:       114 @ bb_check /g/g0/hutter2/post_stage5/deps/AXL/src/axl_async_bbapi.c:134
AXL 0.3.0 ERROR: lassen708: AXL Error with BBAPI details:  {"id":"1","rc":"114","env":{"configfile":"\/etc\/ibm\/bb.cfg"},"error":{"text":"Library was never initialized","func":"BB_TerminateLibrary","line":"296","sourcefile":"\/u\/tgooding\/workspace\/castbuild\/bb\/src\/bbapi.cc","errno":"114","strerror":"Operation already in progress"}} @ bb_check /g/g0/hutter2/post_stage5/deps/AXL/src/axl_async_bbapi.c:135
```
There are a number of ways to get around this, just need to figure out the most elegant one.

tonyhutter:
Fix https://github.com/ECP-VeloC/AXL/pull/94

PR #94, opened 2021-03-02
Don't load the BB API when doing a non-BBAPI transfer

tonyhutter:
Previously, we would blindly call `axl_async_init_bbapi()` and `axl_async_finalize_bbapi()` in AXL_Init/AXL_Finalize if we detected theBBAPI headers.  This became problematic on the IBM login nodes since they had the BBAPI headers, but weren't running the client.

This patch changes the behavior to only load the BBAPI library if you're doing a BBAPI transfer.

Fixes: #70

### DIFF
--- a/src/axl.c
+++ b/src/axl.c
@@ -75,6 +75,10 @@ static unsigned int axl_kvtrees_count = 0;
 
 static pthread_mutex_t id_lock = PTHREAD_MUTEX_INITIALIZER;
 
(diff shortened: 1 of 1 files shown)


adammoody:
Thanks, @tonyhutter .  This PR looks good and is ready to merge.  Though, github is saying it's out of date, so I think it may need to be rebased on master first.

tonyhutter:
Ok, just rebased

PR #101, opened 2021-04-22
Fix BBAPI tag generation

tonyhutter:
The BBAPI requires us to pass a 64-bit tag that will never be repeated on a node.  Previously, we used the unix timestamp +
thread ID for the tag.  This lead to duplicate tags when SCR would start two checkpoints in the same second.  This patch adds in an additional 2-bit quarter second field to the tag along with a quarter second wait, to ensure no identical tags are ever generated on the node.

### DIFF
--- a/src/axl_async_bbapi.c
+++ b/src/axl_async_bbapi.c
@@ -23,6 +23,7 @@
 #include <sys/statfs.h>
 #include <linux/magic.h>
 #include <libgen.h>
+#include <unistd.h>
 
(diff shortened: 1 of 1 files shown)


adammoody:
@tonyhutter .  This looks good to me.  Did you already check that it solves the problem, or are you still waiting on nodes to test?

tonyhutter:
It fixes the issue -  I've been unable to hit the same BB error message since applying the patch.

adammoody:
Thanks, @tonyhutter

PR #131, opened 2022-08-01
WIP - NNFDM implementation in AXL

mcfadden8:
It basically compiles and cannot be fully tested until we have an
operational server.


src/axl_async_nnfdm.c
```diff
@@ -0,0 +1,224 @@
...
+                "NNFDM OpenConnection(%s) failed with error %d @ %s:%d",
+                socketAddr, conn.r1, __FILE__, __LINE__
+            );
+        }
+        nnfdm_connection_ref = conn.r0;
+    }
+}
+
+void axl_async_finalize_nnfdm() {
+    if (nnfdm_initialized) {
```
adammoody, line 32: Another option would be to decrement the reference counter here, and finalize on 0.  I'm not sure if we should actually finalize on the first or last call.

adammoody:
Thanks @mcfadden8 .  Nicely done.

I haven't checked things under this context, but it would be good to think through whether the HPE API supports SCR's scalable restart and scavenge operations.

In the case of a scalable restart, we normally try to cancel any outstanding flush.  Since there is no way to cancel, I think we'd need the restarted job to be able to resume and/or wait on any outstanding flush that was started from a previous run, i.e., I don't think we'd want the restarted job to initiate a new flush of the same files that are already in progress from a flush in a prior run.

For scavenge, is there a way for the job script to see the status of a flush started by the last run?  If not, will there be problems if we try to copy the files again while a flush may still be ongoing?

mcfadden8:
> Thanks @mcfadden8 . Nicely done.
> 
> I haven't checked things under this context, but it would be good to think through whether the HPE API supports SCR's scalable restart and scavenge operations.
> 
> In the case of a scalable restart, we normally try to cancel any outstanding flush. Since there is no way to cancel, I think we'd need the restarted job to be able to resume and/or wait on any outstanding flush that was started from a previous run, i.e., I don't think we'd want the restarted job to initiate a new flush of the same files that are already in progress from a flush in a prior run.
> 
> For scavenge, is there a way for the job script to see the status of a flush started by the last run? If not, will there be problems if we try to copy the files again while a flush may still be ongoing?

Hi @adammoody, for scalable restart, I agree that we would need an API to cancel any  outstanding flushes.

For both restart and scavange, I think we will need a way to list the requests that are still in progress from any previous runs.

I think that these requests may already be documented, but we should discuss to be sure I am understanding things correctly.

src/CMakeLists.txt
```diff
@@ -34,6 +34,10 @@ IF(HAVE_DATAWARP)
     LIST(APPEND libaxl_srcs axl_async_datawarp.c)
 ENDIF(HAVE_DATAWARP)
 
+IF(HAVE_NNFDM)
+    LIST(APPEND libaxl_srcs nnfdm.cpp)
+ENDIF(HAVE_NNFDM)
+
```
adammoody, line 40: I think we'll also want this in ``src/dist/CMakeLists.txt``.

src/nnfdm.cpp
```diff
@@ -0,0 +1,237 @@
...
+    for (kvtree_elem* elem = kvtree_elem_first(files); elem != NULL; elem = kvtree_elem_next(elem)) {
+        int status;
+        char* uid;
+
+        kvtree* elem_hash = kvtree_elem_hash(elem);
+        kvtree_util_get_str(elem_hash, AXL_KEY_FILE_SESSION_UID, &uid);
+
+        do {
+            status = nnfdm_stat(uid, max_seconds_to_wait);
+        } while (status == AXL_STATUS_INPROG);
```
adammoody, line 187: For future work, we may want to add a sleep() in here to avoid thrashing the server will poll requests.

adammoody:
Nice.  Thanks, @mcfadden8

src/axl.c
```diff
@@ -826,6 +847,11 @@ static int __AXL_Add (int id, const char* src, const char* dest)
         break;
 #endif /* HAVE_DATAWARP */
 
+#ifdef HAVE_NNFDM
+    case AXL_XFER_ASYNC_NNFDM:
+        break;
+#endif /* HAVE_NNFDM */
```
adammoody, line 853: Let's include a comment here about why we don't have to do anything.

PR #132, opened 2022-11-01
Add cmake config support

nmm0:
See: https://github.com/ECP-VeloC/KVTree/pull/66


adammoody:
This one looks to be tripping on the non-MPI build.

adammoody:
Dove in a bit.  This looks to be failing when including kvtree, whose cmake config contains a dependency on MPI.

https://github.com/ECP-VeloC/KVTree/blob/3b42c739df4025e6dcbb9deb4480ed226c545294/cmake/kvtreeConfig.cmake.in#L5

I think we'll either need a second file here or somehow make MPI conditional there.

nmm0:
> Dove in a bit. This looks to be failing when including kvtree, whose cmake config contains a dependency on MPI.
> 
> https://github.com/ECP-VeloC/KVTree/blob/3b42c739df4025e6dcbb9deb4480ed226c545294/cmake/kvtreeConfig.cmake.in#L5
> 
> I think we'll either need a second file here or somehow make MPI conditional there.

Ah I see -- yes I think it would need an if (@MPI@) block around the `find_dependency`

adammoody:
Yep, I was able to reproduce the failure, and I can verify that changing the ``kvtreeConfig.cmake`` to the following works:
```
if (@MPI@)
    find_dependency(MPI REQUIRED)
endif (@MPI@)
```
It also seems that cmake 3.14.5 fails to set MPI properties correctly for me.  I get:
```
Could NOT find MPI_C (missing: MPI_C_WORKS)
```
Things work again if I bump to some higher cmake, like 3.16.8.  Maybe we do need to raise that minimum cmake version after all.

adammoody:
> @adammoody I put the patch in a new PR: [ECP-VeloC/KVTree#67](https://github.com/ECP-VeloC/KVTree/pull/67)
> 
> However, I'm not sure about the cmake 3.14 issue. I've built with MPI on my machine, maybe it's related to the cluster? I know CMake has some hardcoded path to search for MPI on some Cray machines...

Don't worry about the CMake version.  We'll figure out what minimum version actually works on our systems and update things appropriately later.  Thanks.

Since MPI is optional for AXL, I'm guessing we also can use a similar fix in the AXL cmake config?

PR #133, opened 2023-02-04
extend matrix to build with and without shared libs

adammoody:
This uses a new ``shared`` setting to toggle shared lib builds of both AXL and its dependencies.  If the shared library is not built for AXL, then its ECP dependencies should not built their shared libs either.

Depends on: https://github.com/ECP-VeloC/github-actions/pull/1

The combo of ``-DBUILD_SHARED_LIBS=OFF`` and ``-DAXL_LINK_STATIC=OFF`` is not expected to work.  We can't link executables to the shared library when it is not built.  We exclude that combo from the matrix:

https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -16,8 +16,13 @@ jobs:
         os: [ubuntu-latest, macos-latest]
         target: [Debug, Release]
         mpi: [seq, par]
-        shared: [true, false]
-        static: [true, false]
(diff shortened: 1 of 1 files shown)


adammoody:
@gonsie , here is the AXL PR that uses the new github actions.

gonsie:
superseded by #136

PR #142, opened 2023-05-30
Improving logic for creating unique ids for BB APIs.

hariharan-devarajan:
The main issue is that with current logic, we have same thread ids across threads. So we now include:

- 16-bit for pid: As we can have different process ids but same thread id.
- 16-bit for tid: As we can have same process id but different thread id.
- 16-bit for timestamp: As in case of restart or failure or pid roleovers we can have same pid and hash of thread.
- 16-bit counter: logical timestamp to keep threads and calls unique.

### DIFF
--- a/src/axl_async_bbapi.c
+++ b/src/axl_async_bbapi.c
@@ -10,6 +10,11 @@
 #include <stdint.h>
 #include "axl_internal.h"
 #include "axl_async_bbapi.h"
+#include <inttypes.h>
+#include <pthread.h>
+#include <limits.h>
+
+/* These aren't always defined in linux/magic.h */
 
 #ifdef HAVE_PTHREADS
 #include "axl_pthread.h"
@@ -26,13 +31,17 @@
 #include <libgen.h>
 #include <unistd.h>
 
-/* These aren't always defined in linux/magic.h */
 #ifndef GPFS_SUPER_MAGIC
 #define GPFS_SUPER_MAGIC 0x47504653 /* "GPFS" in ASCII */
 #endif
 #ifndef XFS_SUPER_MAGIC
 #define XFS_SUPER_MAGIC 0x58465342  /* "XFSB" in ASCII */
 #endif
+
+/* Global varibales */
+
+uint64_t counter=0;
+pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER; 
 
 /* We're either running on a compute node with access to the BBAPI, or
  * we're running on a post-stage job node where we have to spawn bbapi.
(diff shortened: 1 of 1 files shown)


tonyhutter:
You shouldn't need the TID since the counter is going to be unique between threads (due to the lock).   I'm thinking now:
```
22 bit PID (max on Linux) + 18 bit counter + 24 bit timestamp (33 years, to mitigate PID rollover)
```
You'll also need to save the value of counter in the locked section.  

You could try something like this (completely untested!):
```diff
diff --git a/src/axl_async_bbapi.c b/src/axl_async_bbapi.c
index 1203dcd..f18d4aa 100644
--- a/src/axl_async_bbapi.c
+++ b/src/axl_async_bbapi.c
@@ -38,9 +38,9 @@
 #define XFS_SUPER_MAGIC 0x58465342  /* "XFSB" in ASCII */
 #endif
 
-/* Global varibales */
+/* Global variables */
 
-uint16_t counter=0;
+uint32_t counter = 0;
 pthread_mutex_t lock = PTHREAD_MUTEX_INITIALIZER; 
 
 /* We're either running on a compute node with access to the BBAPI, or
@@ -253,28 +253,24 @@ int axl_async_finalize_bbapi(void)
  * Returns 0 on success, 1 otherwise. */
 static BBTAG axl_get_unique_tag(void)
 {
+  uint64_t counter_18;
 
   /* This is somewhat of a hack.
    *
    * We need a 64-bit tag that will never be repeated on this node.  To do
    * that we construct an ID of:
    *
-   * 17-bit (pid) + 15-bit (tid)
-   * 16-bit counter + 16-bit (timestamp)
-   *
+   * 22 bit PID + 18 bit counter + 24 bit timestamp (33 years, to mitigate PID rollover)
    */
   /* Lock counter for shared memory access */
   pthread_mutex_lock(&lock);
-  ++counter; /* To capture call count and inter-thread calls */
+  counter_18 = counter & ((1ULL << 18) - 1); /* To capture call count and inter-thread calls */
+  counter++;
   pthread_mutex_unlock(&lock);
-  const int max_17_bit = USHRT_MAX * 2;
-  const int max_16_bit = USHRT_MAX * 2;
-  const int max_15_bit = USHRT_MAX * 2;
-  uint64_t pid_17 = getpid() % max_17_bit; /* PID is 17 bit*/
-  uint64_t tid_15 = syscall(__NR_gettid) % max_15_bit; /* TID are just hashes so 15 bit should be ok.*/
-  uint64_t timestamp = time(NULL) % max_16_bit; /* To capture restart of apps */
-
-  uint64_t tag = (pid_17 << 47) | (tid_15 << 32) | (timestamp << 16) | counter;
+  uint64_t pid_22 = getpid() &  ((1ULL << 22) - 1);
+  uint64_t timestamp_24 = time(NULL) & ((1ULL << 24) - 1);
+
+  uint64_t tag = (pid_22 << 42) | (counter_18 << 18) | timestamp_24;
   return tag;
 }
 
```

tonyhutter (approved):

